PR #29, opened 2019-03-24
Dbdevelopment

leoneln:
This is the latest... got rid of many not nulls, used camel case, isolated users table and crated additional tables on services, program, intake data, services, and lead.


migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+  addedBy int NOT NULL, /*Foreign key reference to person table */
+  FOREIGN KEY (personId) REFERENCES person(personId),
+  FOREIGN KEY (addedBy) REFERENCES person(personId)
+);
+#DROP TAbLE contact;
+/*Create Demographics Table M:1 Person*/
+CREATE TABLE IF NOT EXISTS demographics (
+  personId int NOT NULL, /*foreing key reference to person table*/
+  demographicId int AUTO_INCREMENT PRIMARY KEY,
+  originCountry varchar(64),
```
joeldenning: It might be better to store the two or three Country Code instead of the full name. The country code is considered more definitive and resilient to changes to a country's name.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+  languageHome varchar(32), /* Primary language spoken at home */
+  englishProficiency varchar(32), /*English Proficiency of Person*/
+  dateUSArrival datetime, /*date of us arrival, day need not be exact and can be defaulted to 1 or 15*/
```
joeldenning: This might make more sense as a `DATE` type instead of `DATETIME`, since we don't care about hours, minutes, or seconds.

Also, when you say the day doesn't need to be exact, are you saying that when we know the month but not the day that we'll just default to the 1st or 15th of the month?

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+  employed varchar(5), /*employed binry Y-U-N*/
```
joeldenning: `employed enum(Yes, No, Unknown)`

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+  weeklyAvgHoursWorked tinyint, /*Can be null if not employed*/
+  emplopymentSector varchar(128), /*Can be null if not employed*/
+  hhSize tinyint,
```
joeldenning: nitpick (not important): I lean more towards spelling out the word `householdSize`, just so it's clear to other people what this means.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+  FOREIGN KEY (addedBy) REFERENCES person(personId)
+);
+#DROP TABLE DEMOGRAPHICS;
+/*Create Address table M:1 with people*/
+CREATE TABLE IF NOT EXISTS address (
+  personId int NOT NULL, /*Foreign key reference to person table*/
+  addressId int AUTO_INCREMENT PRIMARY KEY,
+  address1 varchar(128),
+  address2 varchar(128),
+  duelingType varchar(32), /*Dueling type, home, appartment, duplex, townhome*/
```
joeldenning: I've never heard the term `dueling type` before. Is this supposed to be `dwellingType`?

Also, is this a case for an enum?

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+  addedBy int NOT NULL,
+  FOREIGN KEY (personId) REFERENCES person(personId),
+  FOREIGN KEY (addedBy) REFERENCES person(personId)
+);
+
+/*
+Create users tables: I figured that later on, this application can be scaled to mainain, volunteers, board members, and maybe even donors and staff (HR management) 
+*/
+CREATE TABLE IF NOT EXISTS users (
+    userId int AUTO_INCREMENT PRIMARY KEY,
```
joeldenning: very much nit (not important): the indentation changes here from 2 spaces above this line to 4 spaces on this line and below. Usually it doesn't matter which one is used, as long as it is consistent.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    accessLevel nvarchar(32) NOT NULL, 
+    userName nvarchar(64) NOT NULL,
+    password nvarchar(64) NOT NULL
+);
+
+/*
+Create lead table: Here things get a bit interesting, leads and people are siloed but there is potential for leads to become people, and it would be good to track that..
+*/
+CREATE TABLE IF NOT EXISTS lead (
+	leadId int AUTO_INCREMENT PRIMARY KEY,
```
joeldenning: nit (not important): This has a tab character instead of spaces for indentation. It's usually considered best to not mix tabs and spaces inside of a file.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    lastName varchar(64) NOT NULL,
+    accessLevel nvarchar(32) NOT NULL, 
+    userName nvarchar(64) NOT NULL,
+    password nvarchar(64) NOT NULL
+);
+
+/*
+Create lead table: Here things get a bit interesting, leads and people are siloed but there is potential for leads to become people, and it would be good to track that..
+*/
+CREATE TABLE IF NOT EXISTS lead (
```
joeldenning: Would it make more sense to leave this out of our initial database schema, since we don't have any requirements to build anything for leads yet? And then we add in leads once we know more about what feature we'll need?

When I talked with Luis a couple weeks ago after you left, he indicated that CU's texting database has about 8,000 people that could be considered "leads." But he didn't tell me what information they have about those people or how they would want to record or use leads within the software system.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    gender varchar(32),
+    dateAdded DATETIME DEFAULT CURRENT_TIMESTAMP,
+    addedBy int NOT NULL,
+    FOREIGN KEY (addedBy) REFERENCES person(personId)
+);
+/*
+Create Programs: CU you can define its programs and programs managers in this table.
+*/
+CREATE TABLE IF NOT EXISTS programs (
+	programId int AUTO_INCREMENT PRIMARY KEY,
```
joeldenning: tabs -> spaces

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    zipCode varchar(5) NOT NULL,
+    gender varchar(32),
+    dateAdded DATETIME DEFAULT CURRENT_TIMESTAMP,
+    addedBy int NOT NULL,
+    FOREIGN KEY (addedBy) REFERENCES person(personId)
+);
+/*
+Create Programs: CU you can define its programs and programs managers in this table.
+*/
+CREATE TABLE IF NOT EXISTS programs (
```
joeldenning: Who will insert things into this table? Will it be us, manually? I don't see the list changing super frequently (although I could be wrong), so it might be easier to insert the programs manually instead of having a UI to insert them.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+	programId int AUTO_INCREMENT PRIMARY KEY,
+    programName varchar(32) NOT NULL,
+    description varchar(128),
+    programManager int,
+    createdDate DATETIME DEFAULT CURRENT_TIMESTAMP,
+    programStatus varchar(32),
+    FOREIGN KEY (programManager) REFERENCES person(personId)
```
joeldenning: shouldn't this reference `user(userId)` instead?

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+);
+/*
+Create Services Table: Services are in turn adminstered by a program. Pogram 1:M Services
```
joeldenning: Ah I didn't know the difference between programs and services. Good to know 👍

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+*/
+CREATE TABLE IF NOT EXISTS servcies (
+	serviceId int AUTO_INCREMENT PRIMARY KEY,
```
joeldenning: tabs -> spaces

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    programId INT NOT NULL,
+    serviceName varchar(64),
+    serviceDesc varchar(128),
+    FOREIGN KEY (programId) REFERENCES programs(programId)
+);
+/*
+Create LeadServicesNeeded: When a record is added to leads table it may require more than once serives. This table tracks all services required by each lead.. lead 1:M leadServicesNeeded
+*/
+CREATE TABLE IF NOT EXISTS leadServicesNeeded (
```
joeldenning: This table as well maybe would be better to leave out, until we have specific features that we're going to build for leads? Just so we don't have to predict the future and likely miss something or have irrelevant fields.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    FOREIGN KEY (leadId) REFERENCES lead(leadId)
+);
+
+/*
+FROM HERE ON, THE DATABASE IS GETTING INTO MORE DETAIL AND THIS MAYBE PART OF ADDED MODELS.. FOR DISCUSSION
+Create Person Lead to Client (When a new a lead becomes a client lead first name, 
+last name and gender are turned over to person table, leadToPerson gets an insert to maintain that record)
+leadToClient 1:1 Person
+*/
+CREATE TABLE IF NOT EXISTS leadToClient (
```
joeldenning: This table as well might be best left out until we have a feature to build for leads.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    leadId int NOT NULL,
+    comments varchar(5000),
+    dateLeadToClient date,
+    FOREIGN KEY (personId) REFERENCES person(personId),
+    FOREIGN KEY (leadId) REFERENCES lead(leadId)
+);
+/*
+Create IntakeData Table: Intake data is collected on client inital and subsequent visits to CU I expect this to change quite a bit based on feed back form CU
+*/
+CREATE TABLE IF NOT EXISTS intakeData (
```
joeldenning: This is an interesting table -- talking with Vicky and Luis a couple weeks ago it sounds like the long term hope is to be able to track the time spent with clients. But that we wouldn't start with that.

This table seems to be a combination of tracking visits and capturing one-time information. I think it might make most sense just to capture the one time information for now, with the plan to add more to the database schema once we want to track visits.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+	intakeDataId int AUTO_INCREMENT PRIMARY KEY,
```
joeldenning: tabs -> spaces

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    personId int NOT NULL,
+    volunteering varchar(5), /*Y N*/
+    leadtoCU varchar(64), /*how did person hear of CU, Radi, TV, Newspaper, Friend, N/A current client*/
```
joeldenning: nitpick: the word "lead" is being overloaded here to mean something different than it does everywhere else in this database schema. Thus far, a "lead" has been a person who's not quite a client yet. But now we're using the word "lead" to represent the source of a client. Soft suggestion would be to rename to "clientSource" or something.

Also, this field could be an enum

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    programId int NOT NULL,
```
joeldenning: This is an example of tracking a visit that happened, which is something I think that we want to do but not as a first step.

My understanding is that the front page of the paper intake form captures their personal information, whereas the back page captures the initial visit, including which program/service they enrolled in. And that we're starting with just the front page, deferring visit-specific things to features built later.

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    registeredVoter varchar(5), /*Y N*/
+    dataAdded DATETIME DEFAULT CURRENT_TIMESTAMP,
+    addedBy int NOT NULL,
+    FOREIGN KEY (addedby) REFERENCES person(personId),
+    FOREIGN KEY (personId) REFERENCES person(personId)
+);
+/*
+Create Services Needed Table: all services correspond to a program and a person may require more than on service.
+*/
+CREATE TABLE IF NOT EXISTS PersonServicesNeded (
```
joeldenning: Does this represent services that have been provided to clients? Or only an interest expressed by a client in a service?

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,182 @@
...
+    description varchar(128),
+    programManager int,
+    createdDate DATETIME DEFAULT CURRENT_TIMESTAMP,
+    programStatus varchar(32),
+    FOREIGN KEY (programManager) REFERENCES person(personId)
+);
+/*
+Create Services Table: Services are in turn adminstered by a program. Pogram 1:M Services
+*/
+CREATE TABLE IF NOT EXISTS servcies (
```
joeldenning: Typo: `services`

migrations/sqls/20190323195526-name-of-migration-down.sql
```diff
@@ -1,2 +1,14 @@
 /* Replace with your SQL commands */
-DROP TABLE Dummy;
\ No newline at end of file
+DROP TABLE Dummy;
```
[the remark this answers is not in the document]
joeldenning: Wait I just realized that this is the migration for the dummy table.

We should not modify an existing migration once it's on the master branch. Instead, we should only create new migrations. We should move all of your DROP TABLE statements into `20190324132907-dbdev-down.sql`

backend/server.js
```diff
@@ -1,67 +1,107 @@
...
+    let day = d.getUTCDate();
+    let inserts = [
+      "person",
+      req.body.firstname,
+      req.body.lastname,
+      year,
+      month,
+      day
+    ];
+    let qryString =
```
joeldenning: Small tip: you can use a [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Multi-line_strings) to allow for multiline strings so that you don't have to do the `+=` in order to add to the query.

Example:
```js
const query = `
  SELECT personid, firstname, lastname, date_format(dob, '%m/%d/%Y') as birthDate, gender
  FROM person
  WHERE (firstname = ? OR lastname = ?) AND (year(dob) = ? OR month(dob) = ? OR day(dob) = ?)
`
```

backend/server.js
```diff
@@ -1,67 +1,107 @@
...
+      "SELECT personid,firstname,lastname,date_format(dob,'%m/%d/%Y')as birthDate,gender FROM ?? WHERE ";
+    qryString +=
+      "((firstname = ? OR lastname = ?) AND (year(dob) = ? OR month(dob) = ? OR day(dob) = ?))";
```
joeldenning: This code will report someone named Joel Denning born on May 5, 1980 a duplicate of someone named Joel Denning who was born on December 5, 1990. Which to me doesn't make much sense. My first thought is we should just check the whole date, instead of the month day and year separately? Something like this:

```js
mysql.format('SELECT ... AND dob = ?', [req.body.birthday])
```

backend/server.js
```diff
@@ -1,67 +1,107 @@
...
     if (err) {
-      return databaseError(req, res, err)
+      return databaseError(req, res, err);
+    }
+    let d = new Date(req.body.birthday);
+    let year = d.getUTCFullYear();
+    let month = d.getUTCMonth() + 1;
+    let day = d.getUTCDate();
+    let inserts = [
+      "person",
```
joeldenning: The person table is not user input, so we don't have to worry about SQL injection with it. Since the purpose of [prepared statements]() (the "inserts") is to prevent SQL injection, we actually can just remove this from the inserts array and put `person` directly into the `qryString` string

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,185 @@
...
+CREATE TABLE IF NOT EXISTS users (
+  userId int AUTO_INCREMENT PRIMARY KEY,
+  firstName varchar(64) NOT NULL,
+  lastName varchar(64) NOT NULL,
+  accessLevel ENUM('Administrator','Manager','Staff'), 
+  userName nvarchar(64) NOT NULL,
+  password nvarchar(64) NOT NULL
+);
+INSERT INTO users(firstName,lastName,accessLevel,userName,password)
+VALUES('Leonel','Nieto','Administrator','leonelnieto','password');
```
joeldenning: We shouldn't put this into github -- would give away the password to anyone who looks at this repo. Instead, you can insert into your local and production database your admin user account by following [these instructions in the readme](https://github.com/JustUtahCoders/comunidades-unidas-internal#connecting-to-the-database)

migrations/sqls/20190324132907-dbdev-up.sql
```diff
@@ -0,0 +1,185 @@
...
+  gender ENUM('Male','Female','Non-Binary','Other'),  /*Gender constraint at front end */
+  dateAdded DATETIME DEFAULT CURRENT_TIMESTAMP,
+  dateModified DATETIME DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  addedBy int NOT NULL /*Self Reference to Person ID*/,
+  modifiedBy int NOT NULL /*Self Reference to Person ID*/,
+  FOREIGN KEY (addedBy) REFERENCES person(personId),
+  FOREIGN KEY (modifiedBy) REFERENCES users(userId)
+);
+/*Insert test data into person table*/
+INSERT INTO person(firstname,lastname,dob,gender,addedby,modifiedby) VALUES (
```
joeldenning: Same here -- this shouldn't be in the migration. We should insert this into the DB either manually or via an API request done in the backend server.

PR #42, opened 2019-04-25
Iss37

leoneln:
Implemented api and documented it. A second api was also implemented for listing clients by zip but maybe there is a better way for that. Let me know. Also, I created a view on sql for the list of clients.

### DIFF
--- a/api-docs/list-clients.md
+++ b/api-docs/list-clients.md
@@ -1,0 +1,17 @@
+#list-clients API
+#POST Request
+#URL /api/list-clients/
+#Parameters: {firstName, LastName}
+Return a list of people's first name, last name, date of birth, zip code, phone number added by and added date.
+
+#Result: An array of results
+The result object returns
+-First Name: The first name matching the request
+-Last Name: Last Name matching request
+-dob: date of birth
+-zip: conatct zip code
+-PrimaryPhone: contact phone number
+-addedby: person who added the record to database
+-dateadded: datetime when the record was added to database.
+
+#Result size: the number of people who match a first and last name sent with request.
--- a/api-docs/list-clientsbyzip.md
+++ b/api-docs/list-clientsbyzip.md
@@ -1,0 +1,17 @@
+#list-clientsbyzip API
+#POST Request
+#URL /api/list-clientsbyzip/
+#Parameters: {zip}
+Return a list of people's first name, last name, date of birth, zip code, phone number added by and added date.
+
+#Result: An array of results
+The result object returns
+-First Name: The first name matching the request
+-Last Name: Last Name matching request
+-dob: date of birth
+-zip: conatct zip code
+-PrimaryPhone: contact phone number
+-addedby: person who added the record to database
+-dateadded: datetime when the record was added to database.
+
+#Result size: the number of people who match a zip code sent with request.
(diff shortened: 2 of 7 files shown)


api-docs/list-clients.md
```diff
@@ -0,0 +1,17 @@
+#list-clients API
+#POST Request
+#URL /api/list-clients/
+#Parameters: {firstName, LastName}
```
joeldenning: This documentation doesn't help me understand what I actually need to send in the request. Parameters probably refers to query parameters, but I'm not sure whether to send the parameters with `&` in between them or `,` in between them. I find that the best documentation has code examples. Here's an example of what I mean.

```markdown
## Request
GET /clients?firstName=Mario&lastName=Luigi&zip=84109&page=1

Notes
- The values for the query parameters should be [URL encoded](https://www.w3schools.com/tags/ref_urlencode.asp)
- The `page` query parameter defaults to 1.
- The `zip` query parameter can be a partial zip -- the backend algorithm does a "startsWith" check

## Response
```
```json
{
  "clients": [
    {
      "firstName": "Leonel",
      "lastName": "Nieto",
      "dob": "2019/05/30",
      "zip": "84109-0012",
      "primaryPhone": "8015550888",
      "addedBy": {
        "id": 5,
        "name": "Joel Denning"
      },
      "dateAdded": 123123323222,
    }
  ],
  "pagination": {
    "currentPage": 1,
    "totalPages": 10,
    "pageSize": 100,
    "totalNumClients": 954
  }
}
```
```markdown
Notes
- `dateAdded` is a unix timestamp.
```

api-docs/list-clients.md
```diff
@@ -0,0 +1,17 @@
...
+Return a list of people's first name, last name, date of birth, zip code, phone number added by and added date.
+
+#Result: An array of results
+The result object returns
+-First Name: The first name matching the request
```
joeldenning: this does not tell me the name of the property in the object. See above comment for an example of documentation that I find more useful for people unfamiliar with the data.

api-docs/list-clientsbyzip.md
```diff
@@ -0,0 +1,17 @@
+#list-clientsbyzip API
+#POST Request
+#URL /api/list-clientsbyzip/
```
joeldenning: Instead of a separate url endpoint for this, why not just do this as a query parameter to the `GET /clients` url?

For example: `GET /clients?zip=84109`

api-docs/list-clients.md
```diff
@@ -0,0 +1,17 @@
+#list-clients API
+#POST Request
+#URL /api/list-clients/
+#Parameters: {firstName, LastName}
+Return a list of people's first name, last name, date of birth, zip code, phone number added by and added date.
+
+#Result: An array of results
```
joeldenning: This should be a paginated API, since there will be hundreds/thousands of responses.

api-docs/list-clientsbyzip.md
```diff
@@ -0,0 +1,17 @@
...
+#Parameters: {zip}
+Return a list of people's first name, last name, date of birth, zip code, phone number added by and added date.
+
+#Result: An array of results
+The result object returns
+-First Name: The first name matching the request
+-Last Name: Last Name matching request
+-dob: date of birth
+-zip: conatct zip code
```
joeldenning: typo: `contact`

leoneln:
Thanks for the help with the api docs. I will close this pull request and create q new one with the changes.

Issue #52, opened 2019-05-07
Make the Add Client form be in same order as paper intake form


joeldenning:
See google doc intake form for the order

MeghanBomberger:
Before diving into really changing around the fields wanted to run this by you all. Since the paper form doesn't cleanly follow sections and the goal is for quicker, cleaner data entry experience.

Since the paper form's structure doesn't match the current layout do we want to 
- combine the demographic and contact info sections into a single component and provide a straight shot data entry experience for their data entry people.
or
- have the component "breakpoints" be where the red lines are in the form to tie the user experience on the data entry end together

First option favors speed the second favors accuracy use wise. If their work environment has a lot of potential distractions (frequent interruptions, lots of noise, etc...) I'm leaning towards the second. However if their volume is high the first may be a better route and we can simple add the same red line breaks in the design to tie them visually.

joeldenning:
Oops just saw this now -- we talked a bit in person about this, let me know if there's more here that still needs to be discussed.

PR #54, opened 2019-05-09
corrected a typo and moved e-mail and city/state fields

MeghanBomberger:
Minor changes made in regards to issue #52.
Also adding change to structure proposal to issue notes.

### DIFF
--- a/frontend/add-client/contact-information.component.tsx
+++ b/frontend/add-client/contact-information.component.tsx
@@ -64,18 +64,6 @@ export default function ContactInformation(props: StepComponentProps) {
         </div>
         <div>
           <label>
-            <span>Email</span>
-            <input
-              type="email"
-              value={email}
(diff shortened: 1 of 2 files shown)


frontend/add-client/contact-information.component.tsx
```diff
@@ -89,14 +77,14 @@ export default function ContactInformation(props: StepComponentProps) {
         </div>
         <div>
           <label>
-            <span>State</span>
-            <StateSelect state={state} setState={setState} />
+            <span>City</span>
+            <CityInput state={state} city={city} setCity={setCity} />
```
joeldenning: The reason that the city input was after the state input was so that the autocomplete for cities could autopopulate with only the cities within a certain state.

By moving the city input above it, the city input won't know which state to choose cities from if they follow the form from top to bottom.

PR #58, opened 2019-05-10
Issue 50 01

MeghanBomberger:
updated remaining useState fields to retain input info


frontend/add-client/contact-information.component.tsx
```diff
@@ -89,14 +89,14 @@ export default function ContactInformation(props: StepComponentProps) {
         </div>
         <div>
           <label>
-            <span>State</span>
-            <StateSelect state={state} setState={setState} />
+            <span>City</span>
+            <CityInput state={state} city={city} setCity={setCity} />
```
joeldenning: This change shouldn't be in this PR, since it's relevant to #52, where I commented about how we might not want to do this.

frontend/add-client/demographic-information.component.tsx
```diff
@@ -109,7 +115,22 @@ export default function DemographicInformation(props: StepComponentProps) {
...
-            <span>Are they a eligible to vote?</span>
+            <span>Number household dependents under age 18</span>
+            <input
+              type="number"
+              value={householdSizeChildren}
+              onChange={evt =>
+                setHouseholdSizeChildren(Number(evt.target.value))
+              }
+              required
+              min="0"
```
joeldenning: nitpick (not important): you can actually do `min={0}` and `max={30}`, passing in numbers instead of strings.

This is because dom elements have two types of data: html attributes and javascript properties. The attributes all have to be strings, but the properties can by other data types. And it turns out that react almost always (only aria- and custom elements are exceptions) sets properties, not attributes. And the properties can be numbers instead of strings.

frontend/add-client/demographic-information.component.tsx
```diff
@@ -219,7 +275,7 @@ export default function DemographicInformation(props: StepComponentProps) {
             )}
             <div>
               <label>
-                <span>Pay interval</span>
+                <span>Pay interval frequency</span>
```
joeldenning: Doesn't `interval` already refer to how frequent something happens? So `interval frequency` is redundant?

frontend/add-client/services.component.tsx
```diff
@@ -163,45 +163,47 @@ export default function Services(props: StepComponentProps) {
                 checked={services.financialCoaching}
                 onChange={handleChange}
               />
-              <span>Financial coaching</span>
+              <span>Finacial Coaching</span>
```
joeldenning: typo: `Financial`

frontend/add-client/add-client.component.tsx
```diff
@@ -136,13 +137,14 @@ export type ClientState = {
   registerToVote?: boolean;
   // Client source
   clientSource?: ClientSources | string;
+  otherSource?: string;
```
joeldenning: Remove this

PR #60, opened 2019-05-11
back end api for view clients

leoneln:
Implemented backend api.


backend/apis/clients/list-clients.api.js
```diff
@@ -0,0 +1,25 @@
+const { app, databaseError, pool, invalidRequiest } = require("../../server");
+const mysql = require("mysql");
+
+app.get("/api/clients", (req, res, next) => {
```
joeldenning: Like we discussed on Wednesday morning, let's first start with the API contract markdown file, instead of the backend implementation

leoneln: Sounds good.

PR #61, opened 2019-05-12
view clients api doc

leoneln:


api-docs/list-clients.md
```diff
@@ -0,0 +1,34 @@
...
+```json
+{
+  "numClients": 1,
+  "clients": [
+    {
+      "id": 56,
+      "firstName": "Mario",
+      "lastName": "Luigi",
+      "zip": "84107",
+      "dob": "01/01/1981",
```
joeldenning: In the create-client.md file, we called this `birthday` instead of `dob`. See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/create-client.md#request

api-docs/list-clients.md
```diff
@@ -0,0 +1,34 @@
...
+      "primaryPhone": "5551111111",
+      "dateAdded": "04/28/2019",
+      "addedBy": "Shigeru Miyamoto"
```
joeldenning: See the `createdBy` object in https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/create-client.md#success. I think we should be consistent and return the same object here.

api-docs/list-clients.md
```diff
@@ -0,0 +1,34 @@
...
+  "numClients": 1,
+  "clients": [
+    {
+      "id": 56,
+      "firstName": "Mario",
+      "lastName": "Luigi",
+      "zip": "84107",
+      "dob": "01/01/1981",
+      "primaryPhone": "5551111111",
+      "dateAdded": "04/28/2019",
```
joeldenning: Our APIs should be using the standard ISO format for dates (YYYY/MM/DD), instead of MM/DD/YYYY format of date. This is because code always understands the ISO format, but can get confused between month and day if you do it MM/DD/YYYY.

See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/3d06f711607a47c520a62cc3ec32eeaa88903c81/backend/apis/clients/add-client.api.js#L226

api-docs/list-clients.md
joeldenning: Also see comment below about how date format should be YYYY/MM/DD instead of MM/DD/YYYY for our APIs.

api-docs/list-clients.md
```diff
@@ -0,0 +1,34 @@
+# View a list of clients
+
+API is provided to query a list of clients from the database.
+
+##Request
+GET /api/clients/?firstName=Mario&lastName=Luigi&zip=84107&page=1
```
joeldenning: You can remove the slash after `clients/` -- API endpoints do not need trailing slashes.

api-docs/list-clients.md
```diff
@@ -0,0 +1,34 @@
...
+
+Notes:
+
+- Result is limited to 100 rows ordered by lastName, firstName
```
joeldenning: 👍 nice clarification, I like it.

api-docs/list-clients.md
```diff
@@ -0,0 +1,34 @@
...
+- Values for the query should be URL encoded
+- "page" query parameter defaults to 1
```
joeldenning: 👍

leoneln:
Got it, thanks

PR #62, opened 2019-05-12
Switching to old tables, with one-to-many relationships.

joeldenning:


backend/server.js
```diff
@@ -19,7 +19,8 @@ exports.pool = mysql.createPool({
   user: process.env.RDS_USERNAME || "root",
   password: process.env.RDS_PASSWORD || "password",
   database: process.env.RDS_DB_NAME || "local_db",
-  port: process.env.RDS_PORT || "3306"
+  port: process.env.RDS_PORT || "3306",
+  multipleStatements: true
```
joeldenning: This allows us to do multiple INSERT statements or SELECT calls all in one call to the database. See https://www.npmjs.com/package/mysql#multiple-statement-queries

migrations/sqls/20190508233238-clients-up.sql
```diff
@@ -0,0 +1,89 @@
...
+);
+
+/*
+Create IntakeData Table: Intake data is collected on client inital visits along with person, contact and demographic data*/
+CREATE TABLE IF NOT EXISTS intakeData (
+  id int AUTO_INCREMENT PRIMARY KEY,
+  clientId int NOT NULL,
+  dateOfIntake date NOT NULL, /*Not time stap because it maybe different*/
+  clientSource ENUM('facebook', 'instagram', 'website', 'promotionalMaterial', 'consulate', 'friend', 'previousClient', 'employee', 'sms', 'radio', 'tv', 'other'), /*how did person hear of CU*/
+  registeredVoter varchar(5), /*Y N*/
```
leoneln: It seems that on cu's intake they use this as a leading question to initiate the following question, register to vote. But does it make more sense to have the question be a demographics table field? just a thought..

joeldenning: 👍 will change

PR #63, opened 2019-05-13
Intake data

joeldenning:
This PR is part 2 of 3 for the remaining work on getting the Add Client feature storing everything in the database. We already have contactInformation and demographics being stored in the db, this one adds the intakeServices (CU services that the client is interested in) to the Add Client api.

### DIFF
--- a/api-docs/list-services.md
+++ b/api-docs/list-services.md
@@ -1,0 +1,26 @@
+# List CU Services
+
+## Request
+
+```http
+GET /api/services
+```
+
+## Response
+
+```json
+{
+  "services": [
+    {
+      "id": 1,
+      "serviceName": "Citizenship",
+      "serviceDescription": "Gain United States citizenship"
+    },
+    {
+      "id": 2,
+      "serviceName": "Family Petition",
+      "serviceDescription": "Petition for certain family members to receive either a Green Card, a fiancé(e) visa or a K-3/K-4 visa"
+    }
+  ]
+}
+```
--- a/backend/apis/clients/add-client.api.js
+++ b/backend/apis/clients/add-client.api.js
@@ -164,8 +164,8 @@ app.post("/api/clients", (req, res, next) => {
           clientId,
           req.body.dateOfIntake,
           requestEnum(req.body.clientSource),
-          requestBoolean(req.body.eligibleToVote),
           requestBoolean(req.body.registeredToVote),
+          requestBoolean(req.body.eligibleToVote),
           requestBoolean(req.body.couldVolunteer),
           req.session.passport.user.id
         ];
--- a/backend/apis/clients/add-client.api.js
+++ b/backend/apis/clients/add-client.api.js
@@ -1,8 +1,8 @@
 const { app, databaseError, pool } = require("../../server");
 const mysql = require("mysql");
 const {
-  responseDateWithoutTime,
   requestEnum,
+  requestBoolean,
   requestPhone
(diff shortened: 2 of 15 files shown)


frontend/add-client/services.component.tsx
```diff
@@ -24,177 +36,21 @@ export default function Services(props: StepComponentProps) {
...
-              <input
-                type="checkbox"
-                name="services"
-                value="voterRegistration"
-                checked={services.voterRegistration}
-                onChange={handleChange}
-              />
-              <span>Voter registration</span>
-            </label>
+            {services.map(service => (
```
joeldenning: Instead of hard coding the services in the frontend, we're now fetching them from the `GET /api/services` endpoint

migrations/sqls/20190512164549-services-list-up.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE services MODIFY COLUMN serviceDesc VARCHAR(1028);
+
+INSERT INTO services (
```
joeldenning: Vicky is working on the descriptions for the services -- I'll create another db migration to update the service descriptions once she gives me them.

api-docs/list-services.md
```diff
@@ -0,0 +1,26 @@
+# List CU Services
```
joeldenning: We're exposing the list of services so that the frontend can render a dynamic list of services instead of having to hard code them.

frontend/add-client/services.component.tsx
leoneln: This is cool..

frontend/add-client/services.component.tsx
```diff
@@ -2,11 +2,23 @@ import React, { useState } from "react";
...
 import { useCss } from "kremling";
+import easyFetch from "../util/easy-fetch";
 
 export default function Services(props: StepComponentProps) {
-  const [services, setServices] = useState(defaultServices);
+  const [services, setServices] = useState([]);
+  const [checkedServices, setCheckedServices] = useState([]);
   const scope = useCss(css);
 
+  React.useEffect(() => {
```
leoneln: So this is what a way to use useEffect... What does the AbortController component do? I am assuming is a react component?

joeldenning: The abort controller allows us to cancel the network request and not call setServices if the user navigates to a different page while we're waiting on the network request.

See https://stackoverflow.com/questions/31061838/how-do-i-cancel-an-http-fetch-request

backend/apis/clients/add-client.api.js
```diff
@@ -193,19 +203,69 @@ app.post("/api/clients", (req, res, next) => {
...
+            connection.release();
+
+            res.send({
+              success: true
+            });
+
+            return;
+          }
+
+          const intakeServicesValues = req.body.intakeServices.reduce(
```
leoneln: Me being curious, what is reduce doing?

joeldenning: We want to transform an array of service ids into an array that can be used in the MySQL query as values. Each mysql query needs both the `intakeId` and the `serviceId`, so we need to turn each service id into two values in the array

```js
const intakeId = 423

// Starting value
[1, 2, 3]

// After we reduce
[423, 1, 423, 2, 423, 3]
```

leoneln:
Looks good.

PR #64, opened 2019-05-14
Get /clients/:id and returning client object from POST /clients

joeldenning:


backend/apis/clients/add-client.api.js
```diff
@@ -262,12 +257,23 @@ app.post("/api/clients", (req, res, next) => {
             }
 
             connection.commit();
+            returnTheClient();
+          });
+        });
+
+        function returnTheClient() {
```
joeldenning: The call to create a client is now returning the client object after it succeeds in creating it.

backend/apis/clients/get-client.api.js
```diff
@@ -0,0 +1,150 @@
+const { app, databaseError, pool, invalidRequest } = require("../../server");
+const mysql = require("mysql");
+const { checkValid, validInteger } = require("../utils/validation-utils");
+const { responseFullName } = require("../utils/transform-utils");
+
+app.get("/api/clients/:id", (req, res, next) => {
```
joeldenning: This API will come in handy when we're building the page for a single client where we show them all the client info and then let them modify any of it that they want to.

backend/apis/clients/get-client.api.js
```diff
@@ -0,0 +1,150 @@
...
+          errors: [`Could not find client with id ${req.params.id}`]
+        });
+      }
+    });
+  });
+});
+
+exports.getClientById = getClientById;
+
+function getClientById(connection, clientId, cbk) {
```
joeldenning: This function is reused by `GET /clients/:id` and `POST /clients`

api-docs/README.md
```diff
@@ -26,8 +26,10 @@ This is all you need for to authenticate and authorize.
 
 [Check duplicate clients](/api-docs/check-duplicate-client.md)
 
-[List clients](/api-docs/list-clients.md)
-
 [Create client](/api-docs/create-client.md)
 
+[Get client](/api-docs/get-client.md)
```
joeldenning: New API

api-docs/list-clients.md
```diff
@@ -2,7 +2,8 @@
 
 API is provided to query a list of clients from the database.
 
-##Request
+## Request
```
joeldenning: Github markdown requires a space character between the `##` and the text in order for it to show up as a header. Compare [before](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/list-clients.md) and [after](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/b4f8c7fb3f5477bedf2a04a3801e5ae57fcce72d/api-docs/list-clients.md)

leoneln: Good to know

PR #72, opened 2019-05-15
Beginnings of the View Client page.

joeldenning:
view-client-1

### DIFF
--- a/backend/apis/clients/get-client.api.js
+++ b/backend/apis/clients/get-client.api.js
@@ -9,7 +9,13 @@ app.get("/api/clients/:id", (req, res, next) => {
       return databaseError(req, res, err, connection);
     }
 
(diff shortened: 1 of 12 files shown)


backend/apis/clients/get-client.api.js
```diff
@@ -9,7 +9,13 @@ app.get("/api/clients/:id", (req, res, next) => {
       return databaseError(req, res, err, connection);
```
leoneln: Can the view clients path be the same..(/api/clients) We had originally planned to use encoded url params to query/search clients or get all clients if no params are submitted..

joeldenning: this is a different url, because it has a specific client ID in it

/api/clients - list of clients
/api/clients/:id - one specific client

PR #73, opened 2019-05-15
Iss37api

leoneln:
Initial code, does not pull anything from db...


leoneln:
I hope I did that right..

joeldenning:
Yeah looks like the merge conflict was resolved correctly 👍

PR #77, opened 2019-05-19
api query created

leoneln:
view client list query created

### DIFF
--- a/backend/apis/clients/list-clients.api.js
+++ b/backend/apis/clients/list-clients.api.js
@@ -1,4 +1,4 @@
-const { app, databaseError, pool, invalidRequiest } = require("../../server");
+const { app, databaseError, pool, invalidRequest } = require("../../server");
 const mysql = require("mysql");
 
 app.get("/api/clients", (req, res, next) => {
@@ -6,18 +6,37 @@ app.get("/api/clients", (req, res, next) => {
     if (err) {
       return databaseError(req, res, err);
     }
-    var queryString = "SELECT ? AS result";
 
-    const query = mysql.format(queryString, [req.query.foo]);
+    var queryString =
+      "SELECT cl.id,cl.firstName,cl.lastName,cl.birthday,ct.zip,ct.primaryPhone,cl.addedBy as addedById,us.firstName as addedByFirstName,us.lastname as addedByLastName,cl.dateAdded FROM clients cl JOIN contactInformation ct ON cl.id = ct.clientId JOIN users us ON cl.addedBy = us.Id WHERE (cl.firstName LIKE ? AND cl.lastName LIKE ?) OR ct.zip = ? ORDER BY cl.firstName,cl.lastName DESC LIMIT 100";
+    if (req.query.page) {
+      queryString += "," + parseInt(req.query.page) * 100;
+    }
+    const getClientList = mysql.format(queryString, [
+      req.query.firstName + "%",
+      req.query.lastName + "%",
+      req.query.zip
+    ]);
 
-    connection.query(query, function(err, rows, fields) {
+    connection.query(getClientList, function(err, rows, fields) {
       if (err) {
         return databaseError(req, res, err);
       }
       res.send({
         numClients: rows.length,
         clients: rows.map(row => ({
-          result: row.result
+          id: row.id,
+          firstName: row.firstName,
+          lastName: row.lastName,
+          birthday: row.birthday,
+          zip: row.zip,
+          primaryPhone: row.primaryPhone,
+          addedBy: {
+            addedById: row.addedById,
+            addedByFirstName: row.addedByFirstName,
+            addedByLastName: row.addedByLastName
+          },
+          dateAdded: row.dateAdded
         }))
       });
     });
--- a/backend/apis/clients/list-clients.api.js
+++ b/backend/apis/clients/list-clients.api.js
@@ -22,7 +22,7 @@ app.get("/api/clients", (req, res, next) => {
       FROM 
         clients cl 
       JOIN 
-        contactInformation ct ON cl.id = ct.clientId 
+        (SELECT * FROM contactInformation ORDER BY dateAdded DESC LIMIT 1) ct ON cl.id = ct.clientId 
       JOIN 
         users us ON cl.addedBy = us.id 
       WHERE 
--- a/backend/apis/clients/list-clients.api.js
+++ b/backend/apis/clients/list-clients.api.js
@@ -1,42 +1,23 @@
 const { app, databaseError, pool, invalidRequest } = require("../../server");
 const mysql = require("mysql");
-const {
-  checkValid,
-  nullableValidInteger
-} = require("../utils/validation-utils");
 
 app.get("/api/clients", (req, res, next) => {
-  const validationErrors = checkValid(req.query, nullableValidInteger("page"));
-
-  if (validationErrors.length > 0) {
-    return invalidRequest(res.validationErrors);
-  }
-
   pool.getConnection((err, connection) => {
     if (err) {
       return databaseError(req, res, err);
     }
 
-    let queryString = `SELECT cl.id, cl.firstName, cl.lastName, cl.birthday, ct.zip, ct.primaryPhone, cl.addedBy as addedById,
-        us.firstName as addedByFirstName, us.lastname as addedByLastName, cl.dateAdded 
-      FROM 
-        clients cl 
-      JOIN 
-        contactInformation ct ON cl.id = ct.clientId 
-      JOIN 
-        users us ON cl.addedBy = us.id 
-      WHERE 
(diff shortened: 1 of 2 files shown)


backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,42 @@
...
 
 app.get("/api/clients", (req, res, next) => {
   pool.getConnection((err, connection) => {
     if (err) {
       return databaseError(req, res, err);
     }
-    var queryString = "SELECT ? AS result";
 
-    const query = mysql.format(queryString, [req.query.foo]);
+    var queryString =
```
joeldenning: nitpick - this can be a `const` instead of `var`. The rule of thumb for modern javascript is to start all variables as `const` and then change them to `let` if you need to modify its value. `var` is usually not used as much anymore.

backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,42 @@
...
+      "SELECT cl.id,cl.firstName,cl.lastName,cl.birthday,ct.zip,ct.primaryPhone,cl.addedBy as addedById,us.firstName as addedByFirstName,us.lastname as addedByLastName,cl.dateAdded FROM clients cl JOIN contactInformation ct ON cl.id = ct.clientId JOIN users us ON cl.addedBy = us.Id WHERE (cl.firstName LIKE ? AND cl.lastName LIKE ?) OR ct.zip = ? ORDER BY cl.firstName,cl.lastName DESC LIMIT 100";
```
joeldenning: This query would be easier to understand if the indentation were better. You can use [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to do multiline strings in javascript.

For example:
```js
const query = `
  SELECT cl.id, cl.firstName, cl.birthday, ct.zip, ct.primaryPhone, cl.addedBy as addedBy,
    us.firstName as addedByFirstName, us.lastName as addedByLastName, cl.dateAdded
  FROM
    clients cl
  JOIN
    contactInformation ct ON cl.id = ct.clientId
  JOIN
    users us on cl.addedBy = us.id
  WHERE
    (cl.firstName LIKE ? AND cl.lastName LIKE ?) OR ct.zip = ?
  ORDER BY cl.firstName, cl.lastName DESC LIMIT 100
`
```

In writing that, I noticed that you have `us.Id` instead of `us.id` -- we should fix that.

backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,42 @@
...
+    if (req.query.page) {
```
joeldenning: We should verify that `req.query.page` is a valid number. See [example](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/d61a30598c0ed44eb7b120ecbfd8938e5aee46c5/backend/apis/clients/get-client.api.js#L15-L19).

We should do this validation check *before getting a database connection*, even though the example I linked to does not. This is because the example I linked to forgot to release the connection before returning.

```js
const validationErrors = checkValid(req.params, nullableValidInteger("page"))
```

The `nullableValidInteger` function is not yet implemented, so you'll have to implement it inside of [validation-utils.js](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/d61a30598c0ed44eb7b120ecbfd8938e5aee46c5/backend/apis/utils/validation-utils.js)

backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,42 @@
...
+      queryString += "," + parseInt(req.query.page) * 100;
+    }
+    const getClientList = mysql.format(queryString, [
+      req.query.firstName + "%",
```
joeldenning: Should we do `'%' + req.query.firstName + '%'` so that it's a "contains" instead of "startsWith"? Also, I think we should move the `%` into the query instead of into the params. The params are escaped for SQL injection, but the `%` is part of the query that we do not want escaped.

backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,42 @@
...
-    connection.query(query, function(err, rows, fields) {
+    connection.query(getClientList, function(err, rows, fields) {
       if (err) {
         return databaseError(req, res, err);
       }
       res.send({
         numClients: rows.length,
         clients: rows.map(row => ({
-          result: row.result
+          id: row.id,
```
joeldenning: 👍

backend/apis/clients/list-clients.api.js
leoneln: Thanks, I will update that.

backend/apis/clients/list-clients.api.js
leoneln: Got it..

backend/apis/clients/list-clients.api.js
leoneln: This maybe over my head. check what i did and let me know..

backend/apis/clients/list-clients.api.js
leoneln: Ill do the former. When i tried to put the % into the query the % was escaped and the query failed. Like this (cl.firstName LIKE %\'Leo\'% AND cl.lastName LIKE %\'Nieto\'%)

backend/apis/utils/validation-utils.js
```diff
@@ -26,6 +26,7 @@ exports.nullableNonEmptyString = nullable(_nonEmptyString);
 exports.nonEmptyString = checkDefined(_nonEmptyString);
 
 exports.nullableValidDate = nullable(_validDate);
+exports.nullableValidInteger = nullable(_validInteger);
```
joeldenning: This is already defined on line 59 -- maybe a merge conflict problem? https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/5cfaafb6abb009dd2e9d2cc281c025979c9b9203/backend/apis/utils/validation-utils.js#L59

backend/apis/utils/validation-utils.js
```diff
@@ -94,13 +95,6 @@ function _validId(propertyName) {
   };
 }
 
-function _validPhone(propertyName) {
```
joeldenning: This should not be deleted. Maybe merge conflict problem?

backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,61 @@
-const { app, databaseError, pool, invalidRequiest } = require("../../server");
+const { app, databaseError, pool, invalidRequest } = require("../../server");
 const mysql = require("mysql");
+const {
+  checkValid,
+  nullableValidInteger
+} = require("../utils/validation-utils");
 
 app.get("/api/clients", (req, res, next) => {
+  const validationErrors = checkValid(req.query, nullableValidInteger("page"));
```
joeldenning: 👍 nice usage of `checkValid` - good job figuring it out

backend/apis/clients/list-clients.api.js
```diff
@@ -1,23 +1,61 @@
...
     }
-    var queryString = "SELECT ? AS result";
 
-    const query = mysql.format(queryString, [req.query.foo]);
+    let queryString = `SELECT cl.id, cl.firstName, cl.lastName, cl.birthday, ct.zip, ct.primaryPhone, cl.addedBy as addedById,
+        us.firstName as addedByFirstName, us.lastname as addedByLastName, cl.dateAdded 
+      FROM 
+        clients cl 
+      JOIN 
+        contactInformation ct ON cl.id = ct.clientId
```
joeldenning: There is a bug here -- there can be multiple contactInformation rows for the same client. But we only want the most recent one. So we should select only the most recent row. See [this code](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/apis/clients/get-client.api.js#L76) for an example of how to query only the most recent one.

backend/apis/clients/list-clients.api.js
joeldenning: Yeah looks good, except for the merge conflict issues I called out in other comments.

backend/apis/clients/list-clients.api.js
leoneln: Got it... That make sense.

backend/apis/utils/validation-utils.js
leoneln: Hopefully i fixed this the right way.

backend/apis/clients/list-clients.api.js
joeldenning: This is pretty difficult to read -- a template string with multiple lines would be easier. See related https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/77/files#r285938028

leoneln: I had already fixed this from your earlier feedback but I broke it yesterday..

PR #78, opened 2019-05-21
Fixed side bar css

MeghanBomberger:
Now compatible cross browser. Updated css to be in vw and vh as a test, seems to be working.
Minor issue with the .product-name. Does not seem to be accepting change unless it is put directly inline.


frontend/navbar/sidebar.component.tsx
```diff
@@ -121,7 +123,7 @@ ${mediaMobile} {
   left: 0;
   top: 0;
   height: 100vh;
-  width: 236rem;
+  width: 30vw;
```
joeldenning: This probably is too small on mobile. `vw` will be quite small on a mobile resolution.

The fix to make it work in all browsers is to change `1rem` to mean `10px` instead of `1px`. There is a bug that is chrome specific relataed to setting 1rem to mean 1px, and the workaround is just to change it to not be 1px.

I'd rather do rems than vh/vw in this case, since i'm more confident that will be the desired behavior for both desktop and mobile in this case.

MeghanBomberger: Addressed in reset with compatibility-attempt-2-01

MeghanBomberger:
Fixed in compatibility-attempt-2-01

PR #87, opened 2019-05-29
Finishing view/edit demographics information

joeldenning:


backend/apis/clients/insert-client.utils.js
```diff
@@ -35,3 +35,52 @@ exports.insertContactInformationQuery = function insertContactInformationQuery(
...
+      employmentSector,
+      payInterval,
+      weeklyAvgHoursWorked,
+      householdSize,
+      dependents,
+      civilStatus,
+      householdIncome,
+      registerToVote,
+      registeredVoter,
+      isStudent,
```
MeghanBomberger: Does the service and date of intake (maybe as a last updated date) data need to be included here?

joeldenning: Date of intake is actually part of the `intakeData` table, not the `demographicsInformation`.

As for service, are you referring to `intakeServices`? Those are part of the `intakeData` as well, not part of demographics

PR #90, opened 2019-05-31
Viewing and updating client source is now working

joeldenning:


frontend/view-edit-client/view-edit-intake-info.component.tsx
```diff
@@ -0,0 +1,132 @@
...
+              type="submit"
+              className="primary"
+              disabled={apiStatus.isUpdating}
+            >
+              Update
+            </button>
+          </div>
+        </form>
+      ) : (
+        <article className="view-edit-intake-info" {...scope}>
```
MeghanBomberger: (nitpick) section may be a better fit here

PR #93, opened 2019-05-31
possible solutuion for autocomplete chrome issue

MeghanBomberger:
Changing input fields that we do not want to autocomplete to autoComplete="new-password" did successfully stop the autocomplete and suggest in Chrome without breaking the same in Firefox. 

Ruled out the false, readonly removed on focus, and false input field methods.

### DIFF
--- a/frontend/util/city-input.component.tsx
+++ b/frontend/util/city-input.component.tsx
@@ -48,7 +48,6 @@ export default function CityInput(props) {
         required
         ref={inputRef}
         autoComplete="new-password"
-        autoCapitalize="words"
         onFocus={() => setSelectedIndex(0)}
         onKeyDown={onKeyDown}
       />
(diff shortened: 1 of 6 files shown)


frontend/util/city-input.component.tsx
```diff
@@ -47,7 +47,7 @@ export default function CityInput(props) {
         onChange={evt => props.setCity(evt.target.value)}
         required
         ref={inputRef}
-        autoComplete="off"
+        autoComplete="new-password"
```
joeldenning: should we do this on all the `<form>`s that we have? Demographics, Services, Client Source, etc?

MeghanBomberger: I think I've found all the spots it was needed in the add client section now

PR #94, opened 2019-05-31
 Viewing and updating intake services is now working. 

joeldenning:


backend/apis/clients/get-client.api.js
```diff
@@ -83,21 +83,33 @@ function getClientById(connection, clientId, cbk) {
       ) intake ON intake.clientId = clients.id
       INNER JOIN users created ON created.id = clients.addedBy
       INNER JOIN users modified ON modified.id = clients.modifiedBy
-      WHERE clients.id = ?
+      WHERE clients.id = ?;
+
+    SELECT serviceId, serviceName
```
joeldenning: The `GET /clients/:id` endpoint was actually not returning the intake services correctly. So I first had to fix that.

frontend/add-client/confirm.component.tsx
```diff
@@ -20,7 +20,7 @@ export default function Confirm(props: StepComponentProps) {
           <button
             type="button"
             className="secondary"
-            onClick={() => props.goBack(Step.CLIENT_SOURCE)}
+            onClick={() => props.goBack(Step.SERVICES)}
```
joeldenning: Fixed a random bug where the Confirm page would skip the SERVICES page when you pressed the Back button

frontend/add-client/form-inputs/intake-services-inputs.component.tsx
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+
+export default React.forwardRef<ServicesRef, IntakeServicesInputsProps>(
```
joeldenning: Pulled this component out of services.component so we could reuse this part in the view-edit-intake-services.component

PR #99, opened 2019-06-01
Beginnings of client list.

joeldenning:
client-list-1-2

### DIFF
--- a/frontend/client-list/client-list.component.tsx
+++ b/frontend/client-list/client-list.component.tsx
@@ -1,0 +1,69 @@
+import React from "react";
+import { useFullWidth } from "../navbar/use-full-width.hook";
+import PageHeader from "../page-header.component";
+import ClientsTableToolbar from "./toolbar/clients-table-toolbar.component";
+import ClientsTable from "./table/clients-table.component";
+import easyFetch from "../util/easy-fetch";
+import { SingleClient } from "../view-edit-client/view-client.component";
+import { useQueryParamState } from "../util/use-query-param-state.hook";
+
+const pageSize = 100;
+
+export default function ClientList(props: ClientListProps) {
+  useFullWidth();
+  const [clientApiData, setClientApiData] = React.useState<ClientApiData>({
+    numClients: 150,
+    clients: []
+  });
+  const [fetchingClients, setFetchingClients] = React.useState(true);
+  const [page, setPage] = useQueryParamState("page", "1", Number);
+
+  React.useEffect(() => {
+    const lastPage = Math.ceil(clientApiData.numClients / pageSize);
+
+    if (typeof page !== "number" || isNaN(page)) {
+      setPage(1);
+    } else if (page <= 0) {
+      setPage(1);
+    } else if (page > lastPage) {
+      setPage(lastPage);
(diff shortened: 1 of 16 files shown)


frontend/util/use-query-param-state.hook.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import queryString from "query-string";
+import { identity } from "lodash-es";
+
+export function useQueryParamState(
```
joeldenning: This hook allows us to put the client list search params into the frontend URL's query string. Which allows the user to refresh the page or copy/paste the URL and maintain their search parameters while on the list clients page.

frontend/client-list/client-list.component.tsx
```diff
@@ -0,0 +1,69 @@
...
+import easyFetch from "../util/easy-fetch";
+import { SingleClient } from "../view-edit-client/view-client.component";
+import { useQueryParamState } from "../util/use-query-param-state.hook";
+
+const pageSize = 100;
+
+export default function ClientList(props: ClientListProps) {
+  useFullWidth();
+  const [clientApiData, setClientApiData] = React.useState<ClientApiData>({
+    numClients: 150,
```
joeldenning: Dummy data that will be used until we actually integrate with the API.

frontend/client-list/client-list.component.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (typeof page !== "number" || isNaN(page)) {
+      setPage(1);
+    } else if (page <= 0) {
+      setPage(1);
+    } else if (page > lastPage) {
+      setPage(lastPage);
+    }
+  }, [page, clientApiData]);
+
+  // React.useEffect(() => {
```
joeldenning: Commented out until we're actually integrating with the API, or at least mocking its response with dummy data.

frontend/client-list/toolbar/clients-table-toolbar.component.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { useIsMobile } from "../../util/use-is-mobile.hook";
+import MobileClientsTableToolbar from "./mobile-clients-table-toolbar.component";
+import DesktopTableToolbar from "./desktop-table-toolbar-component";
+
+export default function ClientsTableToolbar(props: ClientsTableToolbarProps) {
+  const isMobile = useIsMobile();
+
+  return isMobile ? (
```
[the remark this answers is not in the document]
leoneln: Will you have to limit how many columns are shown or just how the information is displayed.

frontend/util/use-query-param-state.hook.tsx
leoneln: Great. This will be handy for sure.

frontend/client-list/toolbar/clients-table-toolbar.component.tsx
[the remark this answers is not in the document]
joeldenning: How the information is displayed -- instead of a table, it will be a card for each client that can grow vertically (where table rows cannot)

PR #100, opened 2019-06-01
Updates to list clients API contract.

joeldenning:


api-docs/list-clients.md
```diff
@@ -4,15 +4,15 @@ API is provided to query a list of clients from the database.
 
 ## Request
 
-GET /api/clients?firstName=Mario&lastName=Luigi&zip=84107&page=1
+GET /api/clients?name=Mario%20Luigi&zip=84107&page=1
```
joeldenning: Just searching by `name` will be an easier UX for the frontend, so we can let the user type in "Juan Rodriguez" without having to change to two different inputs to search for it.

api-docs/list-clients.md
```diff
@@ -25,10 +25,13 @@ Notes:
       "firstName": "Mario",
       "lastName": "Luigi",
       "zip": "84107",
-      "birthday": "1981/01/01",
+      "birthday": "1981-01-01",
```
joeldenning: The ISO format that we're using for dates has `-` instead of `/`. See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/apis/utils/transform-utils.js#L1 for how to format the date correctly.

api-docs/list-clients.md
```diff
@@ -25,10 +25,13 @@ Notes:
...
       "primaryPhone": "5551111111",
-      "dateAdded": "2019/05/13",
-      "createdBy": "Shigeru Miyamoto"
+      "dateAdded": "2019-05-13",
+      "createdBy": {
+        "userId": 123,
```
joeldenning: Always nice to have the userId as well so that we can link to the user or fetch more info about them if needed.

api-docs/list-clients.md
leoneln: I agree that will be better..

PR #120, opened 2019-06-05
96 02 fixes 01

MeghanBomberger:


backend/apis/utils/transform-utils.js
```diff
@@ -1,7 +1,9 @@
 exports.responseDateWithoutTime = date => {
-  const isoString = new Date(date).toISOString();
-  return isoString.slice(0, isoString.indexOf("T"));
+  const dateFormat = require("dateformat");
```
joeldenning: We should not be doing `require` in the middle of the function (which is CommonJS), but `import` at the top of this file (which is ESM). See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/beeeb2098431d826e3292198a66624a2c03eac15/frontend/client-list/table/mobile-clients-table.component.tsx#L4 for an example of what I mean.

frontend/add-client/form-inputs/contact-information-inputs.component.tsx
```diff
@@ -160,8 +160,9 @@ export default React.forwardRef(function ContactInformationInputs(
 });
 
 function getTodayAsString() {
-  const date = new Date().toISOString();
-  return date.slice(0, date.indexOf("T"));
+  const dateFormat = require("dateformat");
```
joeldenning: Same here -- we should use `import` instead of `require`. The documentation for dateformat assumes we're running in NodeJS, but we're not and we should use `import`

package.json
```diff
@@ -29,6 +29,7 @@
     "@reach/router": "^1.2.1",
     "@types/jest": "^24.0.13",
     "@types/jquery": "^3.3.29",
+    "@types/node": "^12.0.4",
```
joeldenning: We do not need `@types/node` because we're not going to use `require()`. We should `yarn remove @types/node`

MeghanBomberger: done

frontend/add-client/form-inputs/contact-information-inputs.component.tsx
MeghanBomberger: done

backend/apis/utils/transform-utils.js
MeghanBomberger: since this is on the backend should a change still be made?

joeldenning: Nope good call -- backend uses require(), not import. The normal thing to do is to add all require() calls to the top of the file, but calling require here works just as well. I'll approve and merge.

PR #126, opened 2019-06-07
Clearing all data from the database, except users. For CU DB launch!

joeldenning:
We will merge this around 3-6pm today, which will signify the end of the "trial" week and the beginning of the db for reals.


migrations/20190607164137-clear-test-data.js
```diff
@@ -0,0 +1,59 @@
+"use strict";
```
joeldenning: This file is generated -- feel free to ignore

PR #134, opened 2019-06-09
Temporary fix for issue 125

MeghanBomberger:
Removed onClick to prevent crashing issue until can be resolved fully.

### DIFF
--- a/frontend/util/city-input.component.tsx
+++ b/frontend/util/city-input.component.tsx
@@ -98,9 +98,6 @@ export default function CityInput(props) {
               <button
                 type="button"
                 className="unstyled city-button"
-                onClick={() => {
-                  setCityAndTab(possibleCity.city);
-                }}
                 tabIndex={-1}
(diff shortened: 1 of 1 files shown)


frontend/util/city-input.component.tsx
```diff
@@ -42,6 +42,7 @@ export default function CityInput(props) {
       onBlur={handleBlur}
     >
       <input
+        id="cityInput"
```
joeldenning: This is unrelated to the purpose of this PR and should be removed

frontend/util/city-input.component.tsx
```diff
@@ -56,6 +57,7 @@ export default function CityInput(props) {
   );
 
   function setCityAndTab(city) {
+    console.log(city);
```
joeldenning: We should not commit or push console.log()'s. I'm actually surprised that this passed the eslint check, I thought we had a rule in there that disallows console.log in committed code.

Issue #138, opened 2019-06-18
Remember which Client Record sections are expanded (small ux improvement)


joeldenning:
Would be nice to store in local storage the most recent expanded/collapsed state for the client record so that if people have information that they prefer to view prominently that they can customize that.

Example of it not remembering when you refresh:
![remember-expannded](https://user-images.githubusercontent.com/5524384/59702259-f6835180-91ff-11e9-96ae-5964ff78209b.gif)

PR #144, opened 2019-06-22
possible solution to 138

MeghanBomberger:
For issue #138

### DIFF
--- a/frontend/view-edit-client/client-section.component.tsx
+++ b/frontend/view-edit-client/client-section.component.tsx
@@ -6,14 +6,21 @@ import { LastUpdate } from "./view-client.component";
 
 export default function ClientSection(props: ClientSectionProps) {
   const [expanded, setExpanded] = React.useState(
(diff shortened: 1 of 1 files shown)


frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,23 @@ import AuditSummarySection from "./audit-summary-section.component";
 import { LastUpdate } from "./view-client.component";
 
 export default function ClientSection(props: ClientSectionProps) {
-  const [expanded, setExpanded] = React.useState(true);
+  const [expanded, setExpanded] = React.useState(
+    JSON.parse(localStorage.getItem(props.title)) === null
```
joeldenning: 👍 nice and simple implementation. Let's use a prefix for the local storage key, so it doesn't get confused with any other local storage values that are set now or in the future.

Maybe `cu-clent-section-expanded:Demographics information`?

PR #152, opened 2019-06-25
145 02

MeghanBomberger:
Having a little trouble converting the success components over to a reusable component when it comes to retaining the github issue path and issue id. I wasn't able to find how it's being passed around from the fetch in the original version. 

I know the path="" and exact added in the attributes are wrong. I had to add them to get it to allow me to push it for a draft PR. Was getting snagged on the pre-commit check.

### DIFF
--- a/frontend/record-client-visit/make-suggestion-success.component.tsx
+++ b/frontend/record-client-visit/make-suggestion-success.component.tsx
@@ -1,0 +1,55 @@
+import React from "react";
+import successIconUrl from "../../icons/148705-essential-collection/svg/success.svg";
+import { useCss } from "kremling";
+import PageHeader from "../page-header.component";
+
+export default function MakeSuggestionSuccess(
+  props: MakeSuggestionSuccessProps
+) {
+  const scope = useCss(css);
+
+  return (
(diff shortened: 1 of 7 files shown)


joeldenning:
> Having a little trouble converting the success components over to a reusable component when it comes to retaining the github issue path and issue id.

See [here](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/73cef6b85d3c14f160f571fe3e74dd1ba9e92298/frontend/report-issue/report-issue.component.tsx#L27) and [here](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/frontend/root.component.tsx#L25). Do those show you how? You can also read more in [reach router docs](https://reach.tech/router/api/RouteComponent)

MeghanBomberger:
I was following there, but when I get to passing it to line 18 and 21 of frontend/util/github-issue-success.component.tsx is where I seem to be passing something incorrectly. The internal path is showing up fine, but I'm getting a blank on the string for line 21 where the issue number should be and its not populating in the link the actual github issue so that link isn't redirecting right.

PR #158, opened 2019-07-01
solution to 138

MeghanBomberger:


frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,30 @@ import AuditSummarySection from "./audit-summary-section.component";
 import { LastUpdate } from "./view-client.component";
 
 export default function ClientSection(props: ClientSectionProps) {
-  const [expanded, setExpanded] = React.useState(true);
+  const [expanded, setExpanded] = React.useState(
+    JSON.parse(
```
joeldenning: No need to `JSON.parse()` here -- can just do 

```js
localStorage.getItem(`cu-client-section-expanded:${props.title}`) === null
```

frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,30 @@ import AuditSummarySection from "./audit-summary-section.component";
...
+      localStorage.getItem(`cu-client-section-expanded:${props.title}`)
+    ) === null
+      ? true
+      : JSON.parse(
+          localStorage.getItem(`cu-client-section-expanded:${props.title}`)
+        )
+  );
   const scope = useCss(css);
 
+  const toggleExpandAndStore = title => {
```
joeldenning: nitpick -- no need to pass `title` into this function. Instead, you can just use `props.title` inside of the function

frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,30 @@ import AuditSummarySection from "./audit-summary-section.component";
 import { LastUpdate } from "./view-client.component";
 
 export default function ClientSection(props: ClientSectionProps) {
-  const [expanded, setExpanded] = React.useState(true);
+  const [expanded, setExpanded] = React.useState(
```
joeldenning: Verrrry nitpick (barely worth mentioning) -- we are checking local storage every render, even though we only need to on the first render.

This really really doesn't matter much at all -- the performance cost of checking local storage is minimal. But I thought I would bring it up to give you a chance to learn about [lazy initial state with the useState hook](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state).

All you'd have to do to change this to lazy initial state is the following:
```js
const [expanded, setExpanded] = React.useState(() => localStorage.getItem('...') ? JSON.parse(localStorage.getItem('...')) : true)
```

frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,30 @@ import AuditSummarySection from "./audit-summary-section.component";
...
+    );
+    setExpanded(!expanded);
+  };
+
   return (
     <div className="card padding-0" {...scope}>
       <button
         className="unstyled client-section-header"
-        onClick={() => setExpanded(!expanded)}
+        onClick={() => toggleExpandAndStore(props.title)}
```
joeldenning: This can just be changed to `onClick={toggleExpandAndStore}` if you implement change suggested above about using  `props.title` inside of the toggleExpandAndStore function.

frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,24 @@ import AuditSummarySection from "./audit-summary-section.component";
...
+    );
+    setExpanded(!expanded);
+  };
+
   return (
     <div className="card padding-0" {...scope}>
       <button
         className="unstyled client-section-header"
-        onClick={() => setExpanded(!expanded)}
+        onClick={() => toggleExpandAndStore}
```
joeldenning: This is a bug. It should be `onClick={toggleExpandAndStore}`. Your code here won't work because the `toggleExpandAndStore` function is not being called. If you pass it in as the entire onClick function, React will call it for us. But if you pass in an arrow function as the onClick, you have to call it.

My suggestion earlier was just to avoid creating an arrow function when you don't need it.

frontend/view-edit-client/client-section.component.tsx
```diff
@@ -5,14 +5,24 @@ import AuditSummarySection from "./audit-summary-section.component";
 import { LastUpdate } from "./view-client.component";
 
 export default function ClientSection(props: ClientSectionProps) {
-  const [expanded, setExpanded] = React.useState(true);
+  const [expanded, setExpanded] = React.useState(() =>
+    localStorage.getItem("...") ? JSON.parse(localStorage.getItem("...")) : true
```
joeldenning: `...` is wrong. That was me not typing out the whole local storage key in my previous comment. But you do need to type out the whole local storage string.

PR #161, opened 2019-07-06
UI for "Client History" page

joeldenning:


frontend/page-header.component.tsx
```diff
@@ -32,26 +33,30 @@ type PageHeaderProps = {
   backgroundColor?: string;
   className?: string;
   fullScreen?: boolean;
+  withSecondaryNav?: boolean;
 };
 
 const css = `
 & .page-header {
   height: 10.2rem;
   display: flex;
```
MeghanBomberger: Would it be better to define the flex box in the mediaMobile page-header with the align-items that was moved there?

joeldenning: Well mediaDesktop is also `display: flex`, so by putting the flex here we're making it so we don't have to put `display: flex` in two places.

PR #187, opened 2019-07-19
More progress on global add case note

joeldenning:
This implements adding a case note through the sidebar. The main new functionality is the "Single Client Input Search" which will be reused for adding a client interaction and possibly for other things, too.

global-add-case-note


joeldenning:
Good question -- there isn't a way to do that right now. We should probably add in birthday to the dropdown, too, so that they can.

👍 I'll do that.

joeldenning:
I updated this to show the birthday for clients

PR #207, opened 2019-08-06
progress update and questions

MeghanBomberger:
Am I on the right track?

### DIFF
--- a/migrations/sqls/20190806211241-client-interaction-up.sql
+++ b/migrations/sqls/20190806211241-client-interaction-up.sql
@@ -1,0 +1,58 @@
+CREATE TABLE IF NOT EXISTS clientInteractions (
+    id INT AUTO_INCREMENT PRIMARY KEY,
+    clientId INT NOT NULL,
+    serviceId INT NOT NULL,
+    interactionType ENUM (
+        'inPerson',
+        'byPhone',
+        'workshopTalk',
+        'oneOnOneLightTouch',
+        'consultation'
+    ) NOT NULL,
+    date DATE DEFAULT CURRENT_DATE NOT NULL,
+    title ENUM (
+        'Client interaction created for Preventive Health: Chronic Disease Screenings',
+        'Client interaction created for Nutrition / CRYS / SNAP: SNAP',
+        'Client interaction created for Nutrition / CRYS / SNAP: Nutrition',
+        'Client interaction created for Nutrition / CRYS / SNAP: Grocery Store Tour',
+        'Client interaction created for Nutrition / CRYS / SNAP: Cooking Classes',
+        'Client interaction created for Immigration: Citizenship',
+        'Client interaction created for Immigration: Family Petition',
+        'Client interaction created for Immigration: DACA',
+        'Client interaction created for Financial Education: Financial Coach',
+        'Client interaction created for Financial Education: Financial Education',
+        'Client interaction created for Workers' Rights: Workers' Rights and Safety',
+        'Client interaction created for Community Engagement and Organizing: Leadership Development',
+        'Client interaction created for Community Engagement and Organizing: Youth Groups'
+    ),
+    duration TIME DEFAULT '0:00:00' NOT NULL,
+    location ENUM (
+        'CUOffice',
+        'consulateOffice',
+        'communityEvent'
+    ) NOT NULL,
+    isDeleted BOOLEAN DEFAULT false NOT NULL,
+    dateAdded DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    addedBy int NOT NULL,
+    dateModified DATETIME DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+    modifiedBy int NOT NULL /*Self Reference to User ID*/, 
+    FOREIGN KEY (clientId) REFERENCES clients(id),
+    FOREIGN KEY (addedBy) REFERENCES users(id),
+    FOREIGN KEY (modifiedBy) REFERENCES users(id),
+    FOREIGN KEY (serviceId) REFERENCES services(id)
+);
+
+ALTER TABLE clientLogs
+    MODIFY COLUMN
+        logType enum (
+            'clientCreated',
+            'clientUpdated:basicInformation',
+            'clientUpdated:contactInformation',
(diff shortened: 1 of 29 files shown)


migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE IF NOT EXISTS clientInteractions (
+    id INT AUTO_INCREMENT PRIMARY KEY,
+    clientInteractionType ENUM (
+        'clientInteractionCreated',
+        'clientInteractionUpdated',
+        'clientInteractionDeleted'
+    ) NOT NULL,
+    clientId INT NOT NULL,
+    title VARCHAR(255) NOT NULL,
```
joeldenning: This gets tricky -- I'm not sure exactly what to do here. We already have `title` and `description` columns [here](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/migrations/sqls/20190629230418-activity-log-up.sql#L4). My idea for client interactions is that the data will be split up between the clientLogs table and the clientInteractions table. Anything that is in the clientLogs table will show up in the Client "History" tab automatically, so it's desireable to put information there instead of in clientInteractions.

Regarding the title -- for many of the clientLogs, we just put in a standard title for the logType. For example, see [this line of code](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/9c8ec0fff60ef58fa220283e5abb9c4b925c14c4/backend/apis/clients/insert-client.utils.js#L30) where the title is always "Contact information was updated."

The only exception to this has been for case notes where we allow the user to type in a title for the case note.

So for client interactions we have the choice of either letting the user specify a title or just creating a default title. I actually lean towards using a default title of "Client interaction added for service SNAP".

Your thoughts?

migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,33 @@
...
+    serviceId INT NOT NULL,
+    interactionType ENUM (
+        'in pesron',
+        'by phone',
+        'workshop/talk',
+        'one on one/light touch',
+        'consultation'
+    ) NOT NULL,
+    description MEDIUMTEXT,
```
joeldenning: This column should be removed, for the same reason as described in my comment about the `title` column

migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,33 @@
...
+    date DATE DEFAULT CURRENT_DATE NOT NULL,
+    duration TIME DEFAULT '0:00:00' NOT NULL,
```
joeldenning: Oh cool I didn't realize that MySQL had a `TIME` data type that exactly matches what our API does. 👍

migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,33 @@
...
+    location ENUM (
```
joeldenning: Hmm maybe same thing about camel casing the enums.

migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,33 @@
...
+        'CU Office',
+        'Consulate Office',
+        'Community Event'
+    ) NOT NULL,
+    canModify BOOLEAN DEFAULT false NOT NULL,
```
joeldenning: This doesn't need to be stored in the database, nor does it need to be in the client interactions API.

The canModify field in the client logs API is a boolean that is false for things like "basic information updated" but true for "case note" and "client interaction." We only need it in the client logs api, not in the client interaction api.

See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/9c8ec0fff60ef58fa220283e5abb9c4b925c14c4/backend/apis/clients/client-logs/activity-log.utils.js#L47 for where this is being calculated for the client logs api

migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,33 @@
...
+    isDeleted BOOLEAN DEFAULT false NOT NULL,
+    dateAdded DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    addedBy int NOT NULL,
```
joeldenning: Let's add two more columns for last update. See linked examples of how to implement it:

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/9c8ec0fff60ef58fa220283e5abb9c4b925c14c4/migrations/sqls/20190508233238-clients-up.sql#L9

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/9c8ec0fff60ef58fa220283e5abb9c4b925c14c4/migrations/sqls/20190508233238-clients-up.sql#L11

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/9c8ec0fff60ef58fa220283e5abb9c4b925c14c4/migrations/sqls/20190508233238-clients-up.sql#L13

test.json
```diff
@@ -0,0 +1,74 @@
+{
```
joeldenning: What is this test file for?

migrations/sqls/20190806211241-client-interaction-up.sql
MeghanBomberger: I think the list would work out the best.

migrations/sqls/20190806211241-client-interaction-up.sql
```diff
@@ -0,0 +1,28 @@
+CREATE TABLE IF NOT EXISTS clientInteractions (
+    id INT AUTO_INCREMENT PRIMARY KEY,
+    clientId INT NOT NULL,
+    serviceId INT NOT NULL,
+    interactionType ENUM (
+        'inPesron',
```
joeldenning: Still typo here: `inPerson.

migrations/sqls/20190806211241-client-interaction-up.sql
joeldenning: What list are you referring to?

MeghanBomberger: A* list, sorry. Of default titles.

PR #216, opened 2019-08-15
Create Client Interaction API

MeghanBomberger:


backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,74 @@
...
+        params.duraction,
+        params.location,
+        params.isDeleted,
+        params.createdBy
+      ]
+    );
+  } else {
+    return mysql.format(
+      `
+				INSERT INTO clientLogs (clientId, title, serviceId, interactionType, description, dateOfInteraction, duraction, createdBy, detailId)
```
joeldenning: Why is this function sometimes inserting into clientInteractions and other times inserting into clientLogs?

backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,74 @@
...
+        params.duraction,
+        params.location,
+        params.isDeleted,
+        params.createdBy
+      ]
+    );
+  }
+};
+
+exports.createResponseInteractionObject = function createResponseInteractionObject(
```
joeldenning: 👍 👍

backend/apis/utils/validation-utils.js
```diff
@@ -68,6 +71,13 @@ function _validDate(propertyName) {
       : `Property ${propertyName} must be a string date of format YYYY-MM-DD. Received '${val}'`;
 }
 
+function _validTime(propertyName) {
+  return val =>
+    /^[0-1][0-9]:[0-5][0-9]:[0-5][0-9]$/.test(val) && !isNaN(new Time(val))
```
joeldenning: I don't think that `Time` is a real function. `new Time()` throws an error when I run it.

MeghanBomberger: Would just removing the  "&& !isNaN(new Time(val))" be enough?

backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,74 @@
+const mysql = require("mysql");
+const { responseFullName } = require("../../utils/transform-utils");
+
+const modifiableInteractionTypes = [
+  "clientInteraction:created",
+  "clientInteraction:updated"
+];
+
+exports.insertClientInteractionsQuery = function(params) {
```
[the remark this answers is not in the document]
MeghanBomberger: Removed

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,89 @@
...
+  if (validationErrors.length > 0) {
+    return invalidRequest(res, validationErrors);
+  }
+
+  const user = req.session.passport.user;
+
+  const sql = mysql.format(
+    `
+			INSERT INTO clientInteractions
+			(clientId, serviceId, title, interactionType, description, dataOfInteraction, duration, location, createdBy)
```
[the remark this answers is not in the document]
MeghanBomberger: Fixed

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,89 @@
...
+      user.id
+    ]
+  );
+
+  pool.query(sql, (err, result) => {
+    if (err) {
+      return databaseError(req, res, err);
+    }
+
+    res.send(
```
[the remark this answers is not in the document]
MeghanBomberger: Added

backend/apis/clients/client-interactions/client-interaction-utils.api.js
MeghanBomberger: Was removed as part of above request

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,89 @@
...
+
+  if (validationErrors.length > 0) {
+    return invalidRequest(res, validationErrors);
+  }
+
+  const user = req.session.passport.user;
+
+  const sql = mysql.format(
+    `
+			INSERT INTO clientInteractions
```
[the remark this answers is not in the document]
MeghanBomberger: Found what was causing this. I use 4 space indent. It's scrambling my spacing when I commit, but doesn't if I remember to switch the file to 2 space before committing.

backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,74 @@
...
+  if (params.detailIdIsLastInsertId) {
+    return mysql.format(
+      `
+				INSERT INTO clientInteractions (clientId, title, serviceId, interactionType, description, dateOfInteraction, duraction, createdBy, detailId)
```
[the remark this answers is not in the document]
MeghanBomberger: see below. fixed.

backend/apis/utils/validation-utils.js
joeldenning: No. `Time` is not a thing in javascript at all. You should just get rid of the `&&` clause altogether because that's just a straight up bug because you're trying to call a function that doesn't exist.

image

backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,32 @@
+const mysql = require("mysql");
+const { responseFullName } = require("../../utils/transform-utils");
+
+const modifiableInteractionTypes = [
```
joeldenning: Delete this. This is not what determines `canModify`.

backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,32 @@
...
+) {
+  return {
+    id: log.id,
+    title: log.title,
+    serviceId: log.serviceId,
+    interationType: log.interactionType,
+    description: log.description,
+    dateOfInteraction: log.dataOfDescription,
+    duration: log.duration,
+    canModify: modifiableInteractionTypes.some(
```
joeldenning: This is wrong. You can modify things that you created and can't modify things that you can't. We should be checking the logged in user for this.

backend/apis/clients/client-interactions/client-interaction-utils.api.js
```diff
@@ -0,0 +1,32 @@
...
+exports.createResponseInteractionObject = function createResponseInteractionObject(
+  log
+) {
+  return {
+    id: log.id,
+    title: log.title,
+    serviceId: log.serviceId,
+    interationType: log.interactionType,
+    description: log.description,
+    dateOfInteraction: log.dataOfDescription,
```
joeldenning: typo: `dataOfDescription`.

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+        createdById: user.id,
+        createdByFirstName: user.firstName,
+        createdByLastName: user.lastName,
+        dateAdded: new Date()
+      })
+    );
+  });
+});
+
+app.post("/clients/:clientId/logs", (req, res) => {
```
joeldenning: This logic should go inside of POST /clients/:id/interactions. It should not be a separate endpoint.

The idea is that the client log gets added to **as a side effect of doing other things**. You don't have to call the api directly because it just gets autopopulated by calling `/clients/:id/interactions`.

The SQL for inserting into clientLogs should go *directly into the POST /interactions api*

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+  if (validationErrors.length > 0) {
+    return invalidRequest(res, validationErrors);
+  }
+
+  const user = req.session.passport.user;
+
+  const sql = mysql.format(
+    `
+      INSERT INTO clientInteractions
+      (clientId, serviceId, title, interactionType, description, dataOfInteraction, duration, location, createdBy)
```
joeldenning: typo: `dateOfInteraction`

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+      VALUES
+      (?, ?, ?, ?, ?, ?, ?, ?, ?)
+    `,
+    [
+      req.params.clientId,
+      req.body.title,
+      req.params.serviceId,
+      req.body.interactionType,
+      req.body.description,
+      req.body.dataOfInteraction,
```
joeldenning: typo: `dateOfInteraction`

backend/apis/clients/client-interactions/client-interaction-utils.api.js
MeghanBomberger: I think I am misunderstanding the code on line 47 of activity-log.utils.js

backend/apis/clients/client-interactions/create-client-interaction.api.js
MeghanBomberger: I seem to have a block about typing that correctly.

backend/apis/clients/client-interactions/create-client-interaction.api.js
MeghanBomberger: fixed

backend/apis/clients/client-interactions/create-client-interaction.api.js
MeghanBomberger: moved

backend/apis/clients/client-interactions/client-interaction-utils.api.js
MeghanBomberger: fixed

backend/apis/clients/client-interactions/client-interaction-utils.api.js
MeghanBomberger: deleted

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+      user.id
+    ]
+  );
+
+  pool.query(sql, (err, result) => {
+    if (err) {
+      return databaseError(req, res, err);
+    }
+
+    res.send(
```
joeldenning: You can only call `res.send()` once. But it's being called twice here.

We should not send the response until both SQL queries are finished.

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+      ]
+    );
+
+    pool.query(sql, (err, result) => {
+      if (err) {
+        return databaseError(req, res, err);
+      }
+
+      res.send(
+        createResponseLogObject({
```
joeldenning: We should not send a client log response object.

This is one api that does two things. We can only have one response object and it should be the one specified in the client interactions API docs.

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+        isDeleted: false,
+        createdById: user.id,
+        createdByFirstName: user.firstName,
+        createdByLastName: user.lastName,
+        dateAdded: new Date()
+      })
+    );
+  });
+
+  app.post("/clients/:clientId/logs", (req, res) => {
```
joeldenning: This needs to be deleted. We are not creating a `POST /clients/:id/logs` api. That is not part of the client interactions project. Instead, we are doing a SQL query inside of `POST /clients/:id/interactions` that inserts a log into the DB.

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,143 @@
...
+    ]
+  );
+
+  pool.query(sql, (err, result) => {
+    if (err) {
+      return databaseError(req, res, err);
+    }
+
+    res.send(
+      createResponseIntereactionObject({
```
joeldenning: typo: `createResponseInteractionObject`

backend/apis/clients/client-interactions/client-interaction-utils.api.js
joeldenning: Let's delete the entire `canModify` property from clientInteractions. From the API docs and from the api.

Client interactions are always modifiable. We don't need to send a `true` for people to know that. My earlier comment yesterday was me getting confused. But we just need to delete `canModify` entirely since it's not helpful.

Also, btw, this code you have here would completely die if you ran it, since `modifiableInteractionTypes` is not defined.

backend/apis/clients/client-interactions/create-client-interaction.api.js
MeghanBomberger: Sorry, I don't think I understand exactly how to do this part. How do I send to two parts of the database with one res.send?

MeghanBomberger: Nevermind I think I might have figured it out finally. Testing now.

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,99 @@
...
+} = require("../../utils/validation-utils");
+
+const { createResponseInteractionObject } = require("./interaction-log.utils");
+
+const { createResponseLogObject } = require("./activity-log.utils");
+
+app.post("/clients/:clientId/interactions", (req, res) => {
+  const validationErrors = [
+    ...checkValid(req.params, validId("clientId")),
+    ...checkValid(req.params, validId("serviceId")),
```
joeldenning: `serviceId` is not in the params. The endpoint is `"/clients/:clientId/interactions"` which does not have a `serviceId` in it.

backend/apis/clients/client-interactions/create-client-interaction.api.js
```diff
@@ -0,0 +1,99 @@
...
+        dateOfInteraction: req.body.dateOfInteraction,
+        duration: req.body.duration,
+        location: req.body.location,
+        isDeleted: false,
+        createdById: user.id,
+        createdByFirstName: user.firstName,
+        createdByLastName: user.lastName,
+        dateAdded: new Date()
+      }),
+      createResponseLogObject({
```
joeldenning: This is still not right. Let's talk over slack or even maybe voice call to iron this out.

PR #220, opened 2019-08-17
Changing deletion of client logs to be a soft delete. Resolves #186

### LINKED ISSUE #186: Backend: change delete client log to do a soft delete instead of hard delete.

joeldenning:
See #186


backend/apis/clients/client-logs/get-activity-logs.api.js
```diff
@@ -23,7 +23,7 @@ app.get("/api/clients/:clientId/logs", (req, res, next) => {
       users.id createdById, users.firstName createdByFirstName, users.lastName createdByLastName
     FROM clientLogs JOIN users
     ON clientLogs.addedBy = users.id
-    WHERE clientLogs.clientId = ?
+    WHERE clientLogs.clientId = ? AND isDeleted = false
```
MeghanBomberger: Should this be "true"?

joeldenning: No. The `GET /api/clients/:id/logs` endpoint should only return logs that are **not deleted**

Issue #185, opened 2019-07-13
Frontend: Modify or Delete a client interaction


joeldenning:
This ties into the the kebob/hamburger menu and shows a modal when the user wants to modify or delete a client interaction.

PR #221, opened 2019-08-17
Setting up frontend to modify/delete client logs. See #185.

joeldenning:
See #185. In the below gif, an error pops up because the backend for modifying client logs is not yet implemented.

modify-client-logs

### DIFF
--- a/frontend/util/modal.component.tsx
+++ b/frontend/util/modal.component.tsx
@@ -1,0 +1,126 @@
+import React from "react";
+import { useCss } from "kremling";
+import { mediaDesktop, mediaMobile } from "../styleguide.component";
+
+export default function Modal(props: ModalProps) {
+  const scope = useCss(css);
+  return (
+    <>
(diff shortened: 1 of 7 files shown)


frontend/util/modal.component.tsx
```diff
@@ -0,0 +1,126 @@
...
+export default function Modal(props: ModalProps) {
+  const scope = useCss(css);
+  return (
+    <>
+      <div className="modal-screen" {...scope} />
+      <div className="modal-dialog" {...scope}>
+        <div className="modal-header">
+          <div>{props.headerText}</div>
+          <button className="icon close" onClick={props.close}>
+            &times;
```
MeghanBomberger: Are the & and ; typos?

joeldenning: No, the `&` indicates the beginning of an html encoded entity and the `;` indicates the end of it. See https://www.w3schools.com/html/html_entities.asp

PR #222, opened 2019-08-17
Implement PATCH client logs. For updating case notes.

joeldenning:
This API is only going to be used for updating case notes. Not for updating client interactions.

### DIFF
--- a/backend/apis/clients/client-logs/patch-activity-log.api.js
+++ b/backend/apis/clients/client-logs/patch-activity-log.api.js
@@ -1,0 +1,95 @@
+const { app, pool, invalidRequest, databaseError } = require("../../../server");
+const {
+  checkValid,
+  validId,
+  nullableNonEmptyString
+} = require("../../utils/validation-utils");
+const mysql = require("mysql");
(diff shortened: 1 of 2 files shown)


backend/apis/clients/client-logs/patch-activity-log.api.js
```diff
@@ -0,0 +1,95 @@
...
+        res,
+        `No client log exists for client ${req.params.clientId} and log ${req.params.logId}`
+      );
+    } else {
+      const existingLog = result[0];
+
+      if (existingLog.addedBy !== req.session.passport.user.id) {
+        return invalidRequest(
+          res,
+          `You may not modify a client log that you did not create`
```
MeghanBomberger: Do you think this could cause issues for if a user is no longer with the organization? Don't know if its not something we need to worry about now.

joeldenning: yeah i figure there might come a day when we need "admin" users who can do anything. But that we can build that when the time comes.

Issue #233, opened 2019-08-20
Add caption to case note


joeldenning:
Add caption to case note explaining not to use it for interactions.

PR #236, opened 2019-08-23
instructions added to case note form

MeghanBomberger:
intended to resolve issue #233


frontend/view-edit-client/client-history/client-history.component.tsx
```diff
@@ -33,6 +33,7 @@ export default function ClientHistory(props: ClientHistoryProps) {
               }
             }))
           });
+          console.log(data);
```
joeldenning:
```suggestion
```

PR #276, opened 2019-09-17
Adding integrations route, behind feature toggle

joeldenning:
icons

### DIFF
--- a/frontend/home/home.component.tsx
+++ b/frontend/home/home.component.tsx
@@ -26,14 +26,14 @@ export default function Home(props: HomeProps) {
           link="add-client"
         />
         <HomeCard
-          iconUrl={caseNoteUrl}
-          title="Add a case note"
-          link="add-case-note"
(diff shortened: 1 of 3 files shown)


frontend/home/home.component.tsx
```diff
@@ -25,16 +25,16 @@ export default function Home(props: HomeProps) {
...
-          iconUrl={caseNoteUrl}
-          title="Add a case note"
-          link="add-case-note"
-        />
         <HomeCard
           iconUrl={recordVisitUrl}
           title="Add a client interaction"
           link="add-client-interaction"
         />
+        <HomeCard
```
MeghanBomberger: Just to clarify, why are we moving this card?

joeldenning: Good question -- it's to match the ordering of the features in the sidenav. The homepage and sidenav listed the features in different orders.

PR #277, opened 2019-09-17
draft of leads api docs

MeghanBomberger:
Draft of the leads project to get a head start on that. 

We should see if they need the source of the lead documented. i.e. what event the lead was generated from.

### DIFF
--- a/api-docs/leads.md
+++ b/api-docs/leads.md
@@ -24,9 +24,20 @@ GET /api/leads/:id
   "dateOfSignup": "2019-09-17",
   "leadStatus": "active",
   "contactStage" {
-    "first": true,
-    "second": false,
-    "third": false
+    "first": "2019-05-06T06:00:00.000Z",
+    "second": null,
+    "third": null
+  },
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16",
+    "eventLocation": "Saint Marks",
+    "relatedProgram": {
+      "id": 7,
+      "programName": "Preventative Health",
+      "programDescription": "Preventative Health"
+    }
   },
   "firstName": "Joel",
   "lastName": "Denning",
@@ -64,9 +75,10 @@ GET /api/leads/:id
 **_Notes_**
 
 - `leadStatus` is an enum with the possible values of `active`, `inactive`, and `converted to client`.
+- `eventSource` is an array of integer event ids
 - `gender` is an enum with possible values of `female`, `male`, `transgender`, `nonbinary`, `other`.
 - `intakeServices` is an array of integers service ids. See (/api-docs/list-services.md).
-- `created.timestamp` and `lastUpdated.timestamp` are ISO timestamps.
+- `contactStage.firstAttempt`, `contactStage.secondAttempt`, `contactStage.thirdAttempt`,`created.timestamp` and `lastUpdated.timestamp` are ISO timestamps. `contactStage.firstAttempt`, `contactStage.secondAttempt`, `contactStage.thirdAttempt` are nullable.
 
 ### Not Found
 
@@ -94,9 +106,14 @@ GET /api/leads
   "dateOfSignup": "2019-09-17",
   "leadStatus": "active",
   "contactStage" {
-    "first": true,
-    "second": false,
-    "third": false
+    "first": "2019-05-06T06:00:00.000Z",
+    "second": null,
+    "third": null
+  },
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16"
   },
   "firstName": "Harry",
   "lastName": "Potter",
@@ -133,9 +150,14 @@ GET /api/leads
   "dateOfSignup": "2019-09-17",
   "leadStatus": "active",
   "contactStage" {
-    "first": true,
-    "second": false,
-    "third": false
+    "first": "2019-05-06T06:00:00.000Z",
+    "second": null,
+    "third": null
+  },
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16"
   },
   "firstName": "Hermione",
   "lastName": "Granger",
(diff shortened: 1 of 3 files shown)


api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
+# Leads
+
+The leads document direct encounters with leads.
```
joeldenning: This sentence doesn't make sense. Could you rework it?

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+
+Leads are potiental clients.
+
+Here is the list of the valid lead types:
```
joeldenning: These look like clientLog types, not lead types. Right? The leads objects themselves don't have a `type`.

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+  "id": 1,
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "pending",
+  "firstName": "Joel",
+  "lastName": "Denning",
+  "fullName": "Joel Denning",
+  "phone": "555-555-5555",
+  "smsConsent": true,
+  "zip": "84115",
+  "age": 25,
```
joeldenning: Age changes, birthday doesn't. We should change to birthday (or birth year) instead of age in the API.

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+  "gender": "male",
```
joeldenning: Gender is not asked for in https://docs.google.com/spreadsheets/d/1wGx9pu2Udnn1MEWhas3u-PTe8uwZeStj20ItJ0UDirA/edit#gid=0

I think we should only expect leads to have the information that is already in that form.

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+
+```json
+{
+  "id": 1,
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "pending",
+  "firstName": "Joel",
+  "lastName": "Denning",
+  "fullName": "Joel Denning",
+  "phone": "555-555-5555",
```
joeldenning: Let's remove the dashes. This not only matches the clients API, but also makes it easier for code to know how to format the phone number. If there are dashes, there might also be parentheses and space characters. Better to just get the raw numbers (as a string)

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+### Request
+
+```http
+GET /api/leads
+```
+
+### Response
+
+```json
+({
```
joeldenning: This is invalid JSON -- should be an array:

```suggestion
[{
```

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+    "timestamp": "2019-05-06T06:00:00.000Z"
+  },
+  "lastUpdatedBy": {
+    "userId": 1,
+    "firstName": "Joel",
+    "lastName": "Denning",
+    "fullName": "Joel Denning",
+    "timestamp": "2019-05-06T06:00:00.000Z"
+  }
+})
```
```suggestion
}]
```

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+POST /api/leads
+```
+
+```json
+{
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "pending",
+  "firstName": "Joel",
+  "lastName": "Denning",
+  "fullName": "Ron Weasley",
```
```suggestion
  "fullName": "Joel Denning",
```

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+    "timestamp": "2019-05-06T06:00:00.000Z"
+  }
+},
+{
+  "id": 2,
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "pending",
+  "firstName": "Hermonie",
+  "lastName": "Granger",
+  "fullName": "Hermonie Granger",
```
joeldenning: for the harry potter nazis: it's spelled "Hermione"

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+    "lastName": "Denning",
+    "fullName": "Joel Denning",
+    "timestamp": "2019-05-06T06:00:00.000Z"
+  }
+}
+```
+
+**Notes**
+
+- `created.timestamp` and `lastUpdated.timestamp` are unix timestamps
```
joeldenning: unix timestamps are numbers, not strings. They are always UTC. We do not use them. The format we use is the ISO date time format.

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+
+```http
+PATCH /api/leads/:id
+```
+
+```json
+{
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "pending",
+	"firstName": "Joel",
```
joeldenning: please change tabs -> spaces

api-docs/leads.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Delete a single Lead
+
+### Request
+
+```http
+DELETE /api/leads/:id
+```
+
+```json
```
joeldenning: HTTP `DELETE` requests cannot have a request body (as part of the http spec). This should be deleted.

api-docs/leads.md
MeghanBomberger: The current form only collects the age. I think they grab the birthday during intake only.

api-docs/leads.md
MeghanBomberger: Should it be changed in the create-client.md too then?

api-docs/leads.md
joeldenning: Hmm but putting ages into a database doesn’t make a ton of sense, since time passes and it becomes stale.

The UI could prompt for age and then make a guess for the birth year and send that up to the API? What do you think?

joeldenning:
Looking back at this PR review, I was maybe a bit harsh in my wording. I would prefer you to not submit draft PRs or change the scope of PRs over time, but think I could have said that better. Let me know if you have any questions about what my expectations are for the changes to this PR.

PR #281, opened 2019-09-18
 Handling "broken" status and "last sync" timestamp 

joeldenning:


frontend/view-edit-client/integrations/integrations.component.tsx
```diff
@@ -109,9 +152,14 @@ const css = `
   height: 1.6rem;
 }
 
-.status {
+& .status {
+  margin: 2.4rem 0;
+}
+
+.status > * {
```
MeghanBomberger: What is the > * for here?

joeldenning: The > means “look for a direct child.” The * means “any type of element”

https://www.w3schools.com/cssref/css_selectors.asp

PR #283, opened 2019-09-18
Implementing GET and PATCH integration APIs

joeldenning:


backend/apis/clients/integrations/patch-integration.api.js
```diff
@@ -0,0 +1,97 @@
...
+  if (validationErrors.length > 0) {
+    return invalidRequest(res, validationErrors);
+  }
+
+  const clientId = Number(req.params.clientId);
+  const integrationId = req.params.integrationId;
+
+  getClientById(clientId, (err, client) => {
+    if (err) {
+      return databaseError(rewq, res, err);
```
MeghanBomberger: req

joeldenning: 👍 good catch. I have updated the code to fix this. Please re-review.

PR #290, opened 2019-09-27
events api docs part 1 - start and get one

MeghanBomberger:
Since they wanted to be able to use event info with the leads I'm also putting together the events api-docs to scope that out. We won't need the full events management part on the front end as part of this phase, but it appears we'll need at least part of the backend and a minimal frontend to cover the needs for the leads phase.

This PR gets us started on it.

### DIFF
--- a/api-docs/events.md
+++ b/api-docs/events.md
@@ -22,8 +22,8 @@ GET /api/event/:id
   "eventLocation": "Saint Marks",
   "programId": {
     "id": 7,
-    "programName": "Preventative Health",
-    "programDescription": "Preventative Health"
+    "programName": "Preventive Health",
+    "programDescription": "Preventive Health"
   },
   "totalAttendence": 150,
   "isDeleted": false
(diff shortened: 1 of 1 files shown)


api-docs/events.md
```diff
@@ -0,0 +1,45 @@
+# Events
+
+The list of events will document CU events.
+
+## Get single Event
+
+### Request
+
+```http
+GET /api/event/:id
```
joeldenning: Where will this API be used in the UI?

MeghanBomberger: If I followed the events part of the conversation right eventually it will be an aspect of the scheduling phase, but for this phase they only need to be able to add basic event data so it can be associated with the leads generated from the event. Maybe a modal form on the leads page for now?

api-docs/events.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Success
+
+```json
+{
+  "id": 1,
+  "eventName": "Heath Fair",
+  "eventDate": "2019-09-16",
+  "eventLocation": "Saint Marks",
+  "programId": {
```
joeldenning: Events aren't tied to a program in https://docs.google.com/spreadsheets/d/1wGx9pu2Udnn1MEWhas3u-PTe8uwZeStj20ItJ0UDirA/edit?usp=drive_web&ouid=117418984505592746114

Let's remove this since it's not needed. I don't want to build features for hypothetical futures.

PR #291, opened 2019-09-27
events apidocs part 2 - get all

MeghanBomberger:
adding the get all

part 2 building off of where PR #290 left off


joeldenning:
Closing because we shouldn't have all these PRs open at once. Please only open PRs into the master branch, like we discussed over Slack.

PR #292, opened 2019-09-27
Events api docs part 3 - create

MeghanBomberger:
building off of where PR #291 left off. Adding the create option.


joeldenning:

PR #293, opened 2019-09-27
Events apidocs part 4 - modify

MeghanBomberger:
picking up where PR #292 left off, this adds the modify option.


joeldenning:

PR #294, opened 2019-09-27
Events apidocs part 5 - delete

MeghanBomberger:
Picking up where PR #293 left off, wrapping up with the delete option.


joeldenning:

PR #295, opened 2019-09-27
Leads API docs  part 1 - get one

MeghanBomberger:
Leads docs for phase 4. 

Per our conversation at the CU meeting age and gender were decided to be kept in as is since birth date will be picked up when converting to client replacing the age attribute and both age and gender are needed for event planning and reporting involving leads. 

Incorporated events as the source of the leads per the meeting conversation also.

### DIFF
--- a/api-docs/leads.md
+++ b/api-docs/leads.md
@@ -7,3 +7,85 @@ Here is the list of the valid log types:
 - `leadCreated`,
 - `leadUpdated`,
 - `leadDeleted`
+
+## Get a single Lead
+
+### Request
+
+```http
+GET /api/leads/:id
+```
+
+### Response
+
+```json
+{
+  "id": 1,
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "active",
+  "contactStage" {
+    "first": "2019-05-06T06:00:00.000Z",
+    "second": null,
+    "third": null
+  },
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16",
+    "eventLocation": "Saint Marks",
+    "relatedProgram": {
+      "id": 7,
+      "programName": "Preventative Health",
+      "programDescription": "Preventative Health"
+    }
+  },
+  "firstName": "Joel",
+  "lastName": "Denning",
+  "fullName": "Joel Denning",
+  "phone": "5555555555",
+  "smsConsent": true,
+  "zip": "84115",
+  "age": 25,
+  "gender": "male",
+  "intakeServices": [
+    {
+      "id": 1,
+      "serviceName": "Citizenship",
+      "serviceDescription": "Gain United States citizenship"
+    }
+  ],
+  "isDeleted": false,
+  "createdBy": {
+    "userId": 1,
+    "firstName": "Joel",
+    "lastName": "Denning",
+    "fullName": "Joel Denning",
+    "timestamp": "2019-05-06T06:00:00.000Z"
+  },
+  "lastUpdatedBy": {
+    "userId": 1,
+    "firstName": "Joel",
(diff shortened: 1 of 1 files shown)


api-docs/leads.md
```diff
@@ -7,3 +7,85 @@ Here is the list of the valid log types:
...
+    "third": null
+  },
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16",
+    "eventLocation": "Saint Marks",
+    "relatedProgram": {
+      "id": 7,
+      "programName": "Preventative Health",
```
joeldenning: The name of the program is `Preventive Health`

```suggestion
      "programName": "Preventive Health",
```

api-docs/leads.md
```diff
@@ -7,3 +7,85 @@ Here is the list of the valid log types:
...
+      "programDescription": "Preventative Health"
```
```suggestion
      "programDescription": "Preventive Health"
```

api-docs/leads.md
```diff
@@ -7,3 +7,85 @@ Here is the list of the valid log types:
 - `leadCreated`,
```
joeldenning: I am thinking we should actually get rid of this now -- too much work to create a whole leadLogs database table and then copy things over into the clientLogs table.

We can have a clientLog event for when the lead is converted to a client, but i don't think we should do the intensive audit log stuff for leads that we do for clients.

api-docs/leads.md
```diff
@@ -7,3 +7,85 @@ Here is the list of the valid log types:
...
+
+### Response
+
+```json
+{
+  "id": 1,
+  "dateOfSignup": "2019-09-17",
+  "leadStatus": "active",
+  "contactStage" {
+    "first": "2019-05-06T06:00:00.000Z",
```
joeldenning: I like this approach for tracking both status and timestamps all in the same properties. However, one thing that is missing is if the lead is currently active or not. And possibly _why_ the client is inactive -- whether they said they were not interested or if CU just decided not to continue pursuing them. We should maybe create a different property to capture whether they are active or not?

joeldenning:
Note that this PR is not to master, and is not the first PR for leads docs -- I was confused earlier and shouldn't have reviewed this since it's not the first one.

api-docs/leads.md
```diff
@@ -2,8 +2,86 @@
...
+    "second": null,
+    "third": null
+  },
+  "inactivityReason": null,
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16",
+    "eventLocation": "Saint Marks",
+    "relatedProgram": {
```
joeldenning: Hmm I don't understand the `relatedProgram` thing. It isn't on https://docs.google.com/spreadsheets/d/1wGx9pu2Udnn1MEWhas3u-PTe8uwZeStj20ItJ0UDirA/edit#gid=0 as far as I can tell and isn't something we've talking about.

joeldenning:
Merge conflict - please fix. Once the merge conflict and related programs stuff is fixed, though, I think everything will be good to merge this.

api-docs/leads.md
```diff
@@ -1,3 +1,87 @@
...
+    "second": null,
+    "third": null
+  },
+  "inactivityReason": null,
+  "eventSource": {
+    "eventId": 1,
+    "eventName": "Health Fair",
+    "eventDate": "2019-09-16",
+    "eventLocation": "Saint Marks",
+    "programId": {
```
joeldenning: Events aren't associated to a program in https://docs.google.com/spreadsheets/d/1wGx9pu2Udnn1MEWhas3u-PTe8uwZeStj20ItJ0UDirA/edit#gid=0.

We should remove this.

PR #296, opened 2019-09-27
leads apidocs part 2 - get all

MeghanBomberger:
Adds the get all info. Picks up where PR #295 left off.


joeldenning:

PR #297, opened 2019-09-27
leads api docs part 3 - create

MeghanBomberger:
picks up where pr #296 left off, adds create option


joeldenning:

PR #298, opened 2019-09-27
leads api docs part 0 - starting

MeghanBomberger:


api-docs/leads.md
```diff
@@ -0,0 +1,9 @@
+# Leads
+
+The leads will document potiental clients sourced from CU events.
+
+Here is the list of the valid log types:
+
+- `leadCreated`,
```
joeldenning: This would require the creation of a `leadLogs` table. And then a whole lead audit log UI and stuff. Leads are different from clients and cannot act as if they were clients, including the idea of having a client log.

I don't think we should create a leadLogs table or capture these events.

MeghanBomberger: Will we want to add a "convertedFromLead" to the clientsLog or just the createdClient combined with updating the status in the leads table to cover tracking lead to client conversion? They seemed interested in being able to report on the conversion rates in various ways for their own internal reporting.

MeghanBomberger: Nevermind, you answered what I was asking on the part 1. Sorry about the confusion.

joeldenning: > Will we want to add a "convertedFromLead" to the clientsLog or just the createdClient combined with updating the status in the leads table to cover tracking lead to client conversion?

Yeah we can create some client events when they convert to leads.

PR #299, opened 2019-09-27
leads apidocs part 4 - modify

MeghanBomberger:
picks up where pr #297 left off. adds the modify option


joeldenning:

PR #304, opened 2019-10-01
fix lead apidocs type

MeghanBomberger:
change highlight makes it look bigger than it is. Only the ( to [ on the get all and tabbing over changed

### DIFF
--- a/frontend/add-client/confirm.component.tsx
+++ b/frontend/add-client/confirm.component.tsx
@@ -102,7 +102,7 @@ export default function Confirm(props: StepComponentProps) {
     employmentSector: c.employmentSector,
     payInterval: c.payInterval,
     weeklyEmployedHours: c.weeklyEmployedHours,
-    householdIncome: c.householdIncome || null,
+    householdIncome: c.householdIncome,
     householdSize: c.householdSize,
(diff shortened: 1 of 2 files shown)


frontend/add-client/confirm.component.tsx
```diff
@@ -102,7 +102,7 @@ export default function Confirm(props: StepComponentProps) {
     employmentSector: c.employmentSector,
     payInterval: c.payInterval,
     weeklyEmployedHours: c.weeklyEmployedHours,
-    householdIncome: c.householdIncome,
+    householdIncome: c.householdIncome || null,
```
joeldenning: Why is this here?

MeghanBomberger: Does it not need to be to make it nullable?

joeldenning: I'm just confused why it's in this PR. The pr is for api docs, but this is inside of the frontend code.

MeghanBomberger: Shoot your right. I missed a git stash doing this one. That should only be in the other PR. Sorry about that.

Issue #280, opened 2019-09-18
fix off click on client list bulk action menu


MeghanBomberger:
should close when clicking off of it

PR #306, opened 2019-10-01
fix to bulk action drop down menu

MeghanBomberger:
For issue #280 

Also creates a reusable drop down menu component so we don't have to recreate one when we do the leads front end and reporting front end.

### DIFF
--- a/frontend/client-list/toolbar/clients-table-toolbar.component.tsx
+++ b/frontend/client-list/toolbar/clients-table-toolbar.component.tsx
@@ -3,11 +3,11 @@ import { useCss } from "kremling";
 import easyFetch from "../../util/easy-fetch";
 import backIcon from "../../../icons/148705-essential-collection/svg/back.svg";
 import nextIcon from "../../../icons/148705-essential-collection/svg/next.svg";
-import kabobIcon from "../../../icons/148705-essential-collection/svg/more-1.svg";
 import ClientSearchInput from "../../client-search/client-list/client-search-input.component";
 import { SearchParseValues } from "../../client-search/client-list/client-search-dsl.helpers";
 import { mediaDesktop, mediaMobile } from "../../styleguide.component";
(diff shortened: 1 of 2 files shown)


frontend/util/dropdown-menu-modal.component.tsx
```diff
@@ -0,0 +1,80 @@
...
+          {...scope}
+        />
+      )}
+    </>
+  );
+}
+
+const css = `
+& .kabob-icon {
+	height: 2rem;
```
joeldenning: Please configure your editor globally to use spaces instead of tabs. Also to use 2 spaces instead of 4. This usually doesn't matter because prettier catches it, but because template strings are whitespace sensitive it does matter for this.

https://stackoverflow.com/questions/34174207/how-to-change-indentation-in-visual-studio-code/45671704

frontend/util/dropdown-menu-modal.component.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+& .dropdown-dialog-box {
+	margin-top: 1rem;
+	padding: 1rem;
+	z-index: 1000;
+}
+
+& .dropdown-dialog-box > ul li:hover {
+	background-color: --light-gray
```
joeldenning: CSS requires semicolons.
```suggestion
  background-color: --light-gray;
```

frontend/client-list/toolbar/clients-table-toolbar.component.tsx
```diff
@@ -20,28 +20,14 @@ export default function ClientsTableToolbar(props: ClientsTableToolbarProps) {
...
-                  <ul>
-                    <li onClick={() => openModal("deleting")}>
-                      Delete Selected Clients
-                    </li>
-                  </ul>
-                </div>
-              </div>
-            )}
-          </div>
+          <DropDownMenuModal
```
joeldenning: 👍 good idea creating a reusable component for this

PR #325, opened 2019-10-07
fix to events api docs for missing info

MeghanBomberger:
the createdby and modifiedby were missing in some of the actions.

### DIFF
--- a/api-docs/events.md
+++ b/api-docs/events.md
@@ -119,9 +119,7 @@ POST /api/events
   "eventName": "Job Fair",
   "eventDate": "2019-09-24",
   "eventLocation": "Salt Lake City Library",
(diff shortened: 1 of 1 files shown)


api-docs/events.md
```diff
@@ -91,7 +119,9 @@ POST /api/events
   "eventName": "Job Fair",
   "eventDate": "2019-09-24",
   "eventLocation": "Salt Lake City Library",
-  "totalAttendence": 150
+  "totalAttendence": 150,
+  "createdBy": 1,
```
joeldenning: `createdBy` and `lastUpdatedBy` never need to be included in the request body because they are already included in the session cookie. We should remove them from POST and PATCH request body.

PR #327, opened 2019-10-08
Fix migration filepath

jenseng:
Looks like this might've been gotten renamed and overlooked in the CR that added it?


jenseng:
abandoning in favor of #326, i should've checked first :)

PR #335, opened 2019-10-14
leads api - list leads

MeghanBomberger:
part of issue #308 

also noticed the full name of the lead was left off the get one

### DIFF
--- a/api-docs/leads.md
+++ b/api-docs/leads.md
@@ -85,153 +85,76 @@ If there is no lead with the provided id, you will get a 404 HTTP response, with
 ### Request
 
 ```http
-GET /api/leads
+GET /api/leads?page=12
 ```
+
+**_Notes_**
+
+- Result is limited to 100 rows ordered by lastName and then firstName
+- Values for the query should be URL encoded
+- "page" query parameter defaults to 1
+- If no search terms are provided, the top 100 rows will be returned ordered by lastName, firstName
 
 ### Response
 
 ```json
-[
-  {
-    "id": 1,
-    "dateOfSignUp": "2019-09-17",
-    "leadStatus": "active",
-    "contactStage": {
-      "first": "2019-05-06T06:00:00.000Z",
-      "second": null,
-      "third": null
-    },
-    "inactivityReason": null,
-    "eventSource": {
-      "eventId": 1,
-      "eventName": "Health Fair",
-      "eventLocation": "St Marks Hospital"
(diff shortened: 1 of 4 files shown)


backend/apis/leads/list-leads.api.js
```diff
@@ -0,0 +1,113 @@
...
+      INNER JOIN users created 
+        ON created.id = leads.addedBy
+      INNER JOIN users modified 
+        ON modified.id = leads.modifiedBy
+      INNER JOIN leadServices 
+        ON leadServices.leadId = leads.id
+      INNER JOIN services 
+        ON services.id = leadServices.serviceId
+    WHERE leads.isDeleted = false
+    GROUP BY leadServices.leadId;
```
joeldenning: This SQL query should be paginated. We should also update the [docs for this api](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/leads.md#get-all-leads) to be paginated.

I'm okay with the events api not being paginated because there likely won't be too many of those. But there will be thousands of leads and I think it would be irresponsible to not make the backend and UI paginated.

To make it paginated, see [here](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/apis/clients/list-clients.api.js#L119) and [here](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/apis/clients/list-clients.api.js#L124-L125).

Also, showing the total number of leads that match the search is another nice thing to add to the API. SELECT_FOUND_ROWS() is the way you can do that.

api-docs/leads.md
```diff
@@ -85,153 +85,76 @@ If there is no lead with the provided id, you will get a 404 HTTP response, with
...
-    "smsConsent": true,
-    "zip": "84115",
-    "age": 39,
-    "gender": "male",
-    "leadServices": [
-      {
-        "id": 1,
-        "serviceName": "Citizenship"
+{
+  [
```
joeldenning: This is invalid json (note the red highlight). Might be good to put future api doc json changes into an online json validador such as https://jsonlint.com/ before pushing the code, since there have been several changes recently where you’ve pushed invalid json.

The problem in the json is that you can’t put an array inside of an object without naming the array as a key in the object.

MeghanBomberger:
I caught an error while prepping the front end PR. Was missing the events data in both the get one and the list api.

PR #337, opened 2019-10-21
Converting events and leads from one to one to one to many

MeghanBomberger:


migrations/sqls/20191018202851-multi-event-leads-up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS leadEvents (
+	id INT AUTO_INCREMENT PRIMARY KEY,
```
joeldenning: tabs -> spaces.

Issue #311, opened 2019-10-01
leads list front end


MeghanBomberger:
list view similar to client list

PR #338, opened 2019-10-25
Leads List Base

MeghanBomberger:
Part of issue #311


frontend/leads-list/table/desktop-leads-table.component.tsx
```diff
@@ -0,0 +1,220 @@
...
+                  <td>{lead.id}</td>
+                  <td>{lead.fullName}</td>
+                  <td>{formatPhone(lead.phone)}</td>
+                  <td className="capitalize">
+                    {lead.eventSources[lead.eventSources.length - 1].eventName}
+                  </td>
+                  <td>{dateformat(lead.dateOfSignUp, "m/d/yyyy")}</td>
+                  <td className="status-cell">
+                    {lead.leadStatus === "convertedToClient"
+                      ? `Converted to client - see client #${lead.clientId}`
```
joeldenning: The nested ternaries!!!!! Ha I tweeted about this yesterday: https://twitter.com/Joelbdenning/status/1187149877978886144

Much better in my opinion would be to pull this out into its own function and then use if/else or switch/case:

In the jsx:
```tsx
{showLeadStatus(lead)}
```

Further down in the file:
```tsx
function showLeadStatus(lead) {
  if (lead.leadStatus === "convertedToClient") {
    return "..."
  } else if (lead.leadStatus === "inactive") {
  } else {
  }

  // etc etc etc
}
```

frontend/util/formatters.tsx
```diff
@@ -12,3 +12,11 @@ export function formatPhone(phone) {
     return "";
   }
 }
+
+export function formatDateWithoutTime(timestamp) {
```
joeldenning: What I don't like about this function is that it doesn't always return a value, which means it will implicitly return `undefined` sometimes.

Additionally, this function doesn't appear to be used anywhere. Let's delete it.

migrations/sqls/20191025023019-lead-status-fix-up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE leads ALTER leadStatus SET DEFAULT "active";
```
joeldenning: Do we actually need a db migration for this? Can't this just be something implemented inside of `POST /api/leads`?

frontend/leads-list/table/desktop-leads-table.component.tsx
```diff
@@ -37,7 +37,79 @@ export default function DesktopLeadsTable(props: LeadsTableProps) {
             </tr>
           ) : (
             props.leads.map(lead => {
-              console.log(lead);
+              function determineLeadStatus(
```
joeldenning: small thing, but i think the component would be more readable if this function were pulled out of the jsx entirely -- maybe moved even outside of the component's function entirely. That way the JSX would still be pretty easily readable. Another smaller and less important reason is that when you put a function inside of another function (the arrow function for `props.lead.map(() => {})`), javascript actually has to recreate the function for each item, instead of being able to reuse the function. This is a negligible and not-actually-relevant performance hit. The bigger reason to change it is just for readability of the JSX.

PR #353, opened 2019-10-29
add events api 

MeghanBomberger:
part of issue #309

### DIFF
--- a/backend/apis/clients/add-client.api.js
+++ b/backend/apis/clients/add-client.api.js
@@ -1,6 +1,5 @@
 const { app, databaseError, pool, invalidRequest } = require("../../server");
 const mysql = require("mysql");
-const { requestEnum } = require("../utils/transform-utils");
 const {
   nullableValidDate,
   checkValid,
@@ -99,7 +98,7 @@ app.post("/api/clients", (req, res, next) => {
 
     connection.beginTransaction(err => {
       if (err) {
-        return databaseError(req, ers, err, connection);
+        return databaseError(req, res, err, connection);
       }
 
       const insertClient = mysql.format(
--- a/backend/apis/events/add-event.api.js
+++ b/backend/apis/events/add-event.api.js
@@ -8,23 +8,19 @@ const {
 } = require("../utils/validation-utils");
 
 app.post("/api/events", (req, res, next) => {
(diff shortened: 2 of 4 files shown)


backend/apis/clients/add-client.api.js
```diff
@@ -99,7 +98,7 @@ app.post("/api/clients", (req, res, next) => {
 
     connection.beginTransaction(err => {
       if (err) {
-        return databaseError(req, ers, err, connection);
+        return databaseError(req, res, err, connection);
```
joeldenning: 👍 good catch

backend/apis/events/add-event.api.js
```diff
@@ -0,0 +1,65 @@
...
+      validInteger("totalAttendence")
+    );
+
+    if (validityErrors.length > 0) {
+      return invalidRequest(res, validityErrors, connection);
+    }
+
+    if (err) {
+      return databaseError(req, res, err, connection);
+    }
```
joeldenning: This is a duplicate of lines 12-14.

backend/apis/events/add-event.api.js
```diff
@@ -0,0 +1,65 @@
...
+  validInteger
+} = require("../utils/validation-utils");
+
+app.post("/api/events", (req, res, next) => {
+  pool.getConnection((err, connection) => {
+    if (err) {
+      return databaseError(req, res, err, connection);
+    }
+
+    const validityErrors = checkValid(
```
joeldenning: If the request is invalid, let's not even bother getting a database connection only to immediately release it. In other words, let's move the validity check to be above the `pool.getConnection()` call

backend/apis/events/add-event.api.js
```diff
@@ -0,0 +1,65 @@
...
+    );
+
+    connection.query(newEvent, (newEventErr, result, fields) => {
+      if (err) {
+        return databaseError(req, res, err, connection);
+      }
+
+      const event = result;
+
+      res.json(event);
```
joeldenning: The response object appears to be missing firstName, lastName, fullName, and timestamp (as specified [in the api docs](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/events.md#response))

PR #361, opened 2019-10-30
Create Leads Api

MeghanBomberger:
Part of issue #350 

I need to work on the response to make it match the api-docs, but that's going to take a little reworking of the get-lead.api since it wasn't set up to return multiple and should probably be a separate PR. This does send the data to the database correctly though as far as my testing showed.


backend/apis/leads/add-leads.api.js
```diff
@@ -0,0 +1,112 @@
...
+  validBoolean,
+  validDate,
+  validInteger
+} = require("../utils/validation-utils");
+
+app.post("/api/leads", (req, res) => {
+  const leads = req.body.data;
+  const user = req.session.passport.user;
+
+  for (let i = 0; i < leads.length; i++) {
```
joeldenning: this will throw an error when the request does not have a body (`Cannot read property length of undefined`). Also when the request body is not an array.

To fix, add this before your for loop:

```js
if (!Array.isArray(leads)) {
  return invalidRequest(res, [`POST /api/leads must be called with an array of leads in the request body`])
}
```

backend/apis/leads/add-leads.api.js
```diff
@@ -0,0 +1,112 @@
...
+  pool.query(newLeadsData, (err, result) => {
+    if (err) {
+      return databaseError(req, res, err);
+    }
+
+    const mapLeadIds = result.map(data => {
+      return data.insertId;
+    });
+
+    res.json({ message: `Created leads ${mapLeadIds}.` });
```
joeldenning: `mapLeadIds` is an array, but you're coercing it into a string by using it inside of a template string. Better would be this, since it doesn't rely on type coercion:

```js
const leadIds = result.map(data => data.insertId).join(", ")

res.send({message: `Created leads ${mapLeadIds}`})
```

Also you can use `res.send()` which is a bit different from res.json() in that it sets the http status correctly (iirc)

backend/apis/leads/get-lead.api.js
```diff
@@ -97,7 +97,7 @@ function getLeadById(leadId, cbk, connection) {
           ON leadEvents.leadId = leads.id
         INNER JOIN events
           ON events.id = leadEvents.eventId
-      WHERE leads.id = ? AND leads.isDeleted = false;
+      WHERE leads.id IN (?) AND leads.isDeleted = false;
```
joeldenning: Why did this change? You're allowing for leadId to be an array of leads now? But we're still coercing it to a number on line 41.

joeldenning:
Also please fix merge conflict

backend/apis/leads/add-leads.api.js
```diff
@@ -0,0 +1,116 @@
...
+        for (let j = 1; j < leadEvents.length; j++) {
+          const eventId = leadEvents[j];
+          leadEventsQuery = leadEventsQuery + ", (@leadId, ?)";
+          leadDataArray.push(eventId);
+        }
+      }
+    }
+
+    const newLeadQuery = `${addToLeadQuery}; ${setLeadId}; ${leadServicesQuery}; ${leadEventsQuery};`;
+    leadQuery = `${leadQuery} ${newLeadQuery}`;
```
joeldenning: From what I can tell, `leadQuery` always is an empty string at this point. Do we even need to have both `newLeadQuery` and `leadQuery` variables? Would something like this work?

```js
leadQuery = `${addToLeadQuery}; ${setLeadId}; ${leadServicesQuery}; ${leadEventsQuery};`
```

MeghanBomberger: Wouldn't that only make the query for a single lead though? leadQuery is the entire query together for all the leads and newLeadQuery is the one for the single lead built within the loop and then added to leadQuery at the end. Should I be doing:

`leadQuery = `${leadQuery} ${addToLeadQuery}; ${setLeadId}; ${leadServicesQuery}; ${leadEventsQuery};``

joeldenning: Ah I see! Yeah I missed that this was inside of the for loop. Disregard my previous comment 👍

PR #366, opened 2019-11-04
backend leads search start

MeghanBomberger:
part of issue #352


backend/apis/leads/list-leads.api.js
```diff
@@ -29,6 +29,14 @@ app.get("/api/leads", (req, res, next) => {
     );
   }
 
+  let whereClause = `WHERE leads.isDeleted = false`;
+  let whereClauseValues = [];
+
+  if (req.query.name) {
```
joeldenning: Let's update [the api documentation](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/leads.md) to have the `name` query parameter there.

backend/apis/leads/list-leads.api.js
```diff
@@ -29,6 +29,14 @@ app.get("/api/leads", (req, res, next) => {
...
+    whereClause += `AND CONCAT(leads.firstName, ' ', leads.lastName) LIKE ? `;
```
joeldenning: don't we need a space character at the beginning of the string here, so that `false` on line 32 doesn't mash up against `AND` on line 36?

api-docs/list-leads.md
```diff
@@ -0,0 +1,49 @@
+# View a list of leads
```
joeldenning: Can't this documentation just be merged into https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/leads.md#get-all-leads instead of having its own file?

api-docs/leads.md
```diff
@@ -86,152 +86,43 @@ If there is no lead with the provided id, you will get a 404 HTTP response, with
 
 ## Get all Leads
 
-### Request
+## Request
```
joeldenning: The Request is a sub header of the Get all Leads header and should have one more level of heading because of that.

```suggestion
### Request
```

api-docs/leads.md
```diff
@@ -86,152 +86,43 @@ If there is no lead with the provided id, you will get a 404 HTTP response, with
...
 
 ```http
-GET /api/leads?page=12
+GET /api/leads?name=John%Jane&page=1
```
joeldenning: What is the `%` here?

api-docs/leads.md
```diff
@@ -86,152 +86,43 @@ If there is no lead with the provided id, you will get a 404 HTTP response, with
...
 
-- Result is limited to 100 rows ordered by lastName and then firstName
 - Values for the query should be URL encoded
 - "page" query parameter defaults to 1
-- If no search terms are provided, the top 100 rows will be returned ordered by lastName, firstName
+- "name" can be partial
+- If no search terms are provided, the top 100 rows will be returned
 
-### Response
+## Response
```
```suggestion
### Response
```

api-docs/leads.md
joeldenning:
```suggestion
GET /api/leads?name=John&page=1
```

PR #367, opened 2019-11-04
Beginnings of Add Leads

joeldenning:
This PR is the beginning of the UI for the Add Leads feature. I took @jenseng's commit and adjusted it to set us up for adding events, too.

### DIFF
--- a/frontend/add-leads/add-leads-table.component.tsx
+++ b/frontend/add-leads/add-leads-table.component.tsx
@@ -1,0 +1,74 @@
+import React from "react";
+import AddLeadRow from "./add-lead-row.component";
+import { useCss } from "kremling";
+
+export default function AddLeadsTable({
+  leads,
+  deleteLead,
+  updateLead
+}: AddLeadsTableProps) {
+  const scope = useCss(css);
+
+  return (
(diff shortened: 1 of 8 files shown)


frontend/add-leads/add-leads-table.component.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  & tr input, & tr select {
+    background: rgba(255, 255, 255, 0.75);
+    border: 1px solid rgb(238, 238, 238, 0.75);
+  }
+
+  & tr:focus-within input, & tr:hover input, & tr:focus-within select, & tr:hover select {
+    background: white;
+    border: 1px solid rgb(238, 238, 238);
```
MeghanBomberger: This color is really close to the --very-light-grey we have in the styleguide. `#F3F3F3` is `rgb(239,239,239)`. Should we use that variable here or does it matter?

joeldenning: Good catch I’ll switch to the styleguide color

joeldenning:
I updated this with your feedback

PR #377, opened 2019-11-08
Lead Search

MeghanBomberger:
For issue #352


frontend/leads-search/lead-list/lead-search-dsl.helpers.tsx
```diff
@@ -0,0 +1,101 @@
+import queryString from "query-string";
```
joeldenning: Instead of copy/pasting this entire file and test, let's reuse the client-search-dsl.helpers.tsx.

The functions inside of that file are almost fully reusable, with the only thing needing to change being the allowedSearchFields. Let's allow the allowedSearchFields to be passed in as an argument to the `parseSearch` and `serializeSearch`.

The reason I think that this one is worth consolidating into one file is that it's not a concrete react component that would get more complex as it has to handle more things. It's a file that is already made to be reused, but just needs slight adjustments for the reuse to happen.

frontend/leads-search/lead-list/lead-search-input.component.tsx
```diff
@@ -0,0 +1,305 @@
...
+  );
+  const inputRef = React.useRef<HTMLInputElement>(null);
+  const [programs, setPrograms] = React.useState([]);
+  const [events, setEvents] = React.useState([]);
+
+  React.useEffect(() => {
+    inputRef.current.setCustomValidity(search.parseResult.errors.join(". "));
+  }, [search]);
+
+  React.useEffect(() => {
```
joeldenning: React effects should return a function that cancels the effect. In this case that should be `() => abortController.abort()`

Additionally, it's better practice to not have a "mount effect" that is trying to be the same thing as componentDidMount. Instead, you should treat each api request / effect as a separate effect. In other words, this effect should be split into two - one for fetching services and one for fetching events.

frontend/leads-list/toolbar/lead-search-input.component.tsx
```diff
@@ -0,0 +1,320 @@
...
+        setPrograms(json.programs);
+      })
+      .catch(err => {
+        setTimeout(() => {
+          throw err;
+        });
+      });
+  }, []);
+
+  React.useEffect(() => {
```
joeldenning: This effect should return a function that aborts the network request

frontend/util/list-search/search-dsl.helpers.tsx
```diff
@@ -1,13 +1,13 @@
...
-  zip: "ZIP Code",
-  phone: "Phone",
-  program: "Interest in Program"
-};
+export let allowedSearchFields = {};
+
+let allowedKeys = Object.keys(allowedSearchFields);
 
-const allowedKeys = Object.keys(allowedSearchFields);
+export function setAllowedSearchFields(searchFields) {
```
joeldenning: This has a bug: it does not allow for two react components to simultaneously use the search dsl helpers. Their react effects would call setAllowedSearchFields at different times, but not everytime before calling `parseSearch`, which would mean that sometimes the wrong searchFields would apply.

Instead, let's add the searchFields as an argument to the `parseSearch`, `serializeSearch`, and `deserializeSearch` functions.

The alternative would be something like this, which I think is a more substantial refactor than is worth it:

```js
// each call to dslSearcher returns a searcher that is bound to those allowedSearchFields.
const dslSearcher = new DslSearcher(allowedSearchFields)
dslSearcher.parseSearch()
```

PR #378, opened 2019-11-09
fix to get one lead api

MeghanBomberger:


backend/apis/leads/get-lead.api.js
```diff
@@ -111,23 +111,23 @@ function getLeadById(leadId, cbk, connection) {
...
 
     const lead = {
       id: l.leadId,
       dateOfSignUp: responseDateWithoutTime(l.dateOfSignUp),
       leadStatus: l.leadStatus === null ? "active" : l.leadStatus,
       contactStage: {
-        first: l.firstContactAttempt,
-        second: l.secondContactAttempt,
-        third: l.thirdContactAttempt
+        first: responseDateWithoutTime(l.firstContactAttempt),
```
joeldenning: The api docs indicate that these timestamps should include the time. (See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/leads.md#response).

I think it would be useful to include the timestamp so that we can show what time of day a contact attempt was made. We should either just return the full timestamp or we should update the api docs to reflect what the api actually responds with.

backend/apis/leads/get-lead.api.js
```diff
@@ -111,23 +111,23 @@ function getLeadById(leadId, cbk, connection) {
...
+        second: responseDateWithoutTime(l.secondContactAttempt),
+        third: responseDateWithoutTime(l.thirdContactAttempt)
       },
       inactivityReason: l.inactivityReason,
       eventSources: eventSources.map(event => ({
         eventId: event.eventId,
         eventName: event.eventName,
         eventLocation: event.eventLocation,
-        eventDate: event.eventDate
+        eventDate: responseDateWithoutTime(event.eventDate)
```
joeldenning: 👍 this matches the API

PR #380, opened 2019-11-09
Fixing typo and bugs in POST /api/events

joeldenning:
attendance was misspelled everywhere. Also, there were two bugs in the POST /api/events api.

### DIFF
--- a/backend/apis/events/add-event.api.js
+++ b/backend/apis/events/add-event.api.js
@@ -14,11 +14,11 @@ app.post("/api/events", (req, res) => {
     nonEmptyString("eventName"),
     nonEmptyString("eventLocation"),
     validDate("eventDate"),
-    validInteger("totalAttendence")
+    validInteger("totalAttendance")
   );
 
   if (validityErrors.length > 0) {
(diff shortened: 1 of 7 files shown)


backend/apis/events/add-event.api.js
```diff
@@ -14,11 +14,11 @@ app.post("/api/events", (req, res) => {
     nonEmptyString("eventName"),
     nonEmptyString("eventLocation"),
     validDate("eventDate"),
-    validInteger("totalAttendence")
+    validInteger("totalAttendance")
   );
 
   if (validityErrors.length > 0) {
-    return invalidRequest(res, validityErrors, connection);
```
joeldenning: `connection` is not defined. The server was dying whenever there was an invalid request to POST /api/events.

backend/apis/events/add-event.api.js
```diff
@@ -52,7 +52,7 @@ app.post("/api/events", (req, res) => {
 
     getEventById(result.insertId, (err, event) => {
       if (err) {
-        return err(req, res);
```
joeldenning: `err` is not a function. This caused the server to die.

PR #382, opened 2019-11-09
Implementing the ui for creating events.

joeldenning:
image

image

### DIFF
--- a/frontend/add-leads/add-event-step.component.tsx
+++ b/frontend/add-leads/add-event-step.component.tsx
@@ -51,7 +51,7 @@ export default function AddEventStep(props: AddEventStepProps) {
         }
       })
         .then(result => {
-          console.log("result", result);
+          // TO-DO: move to the add leads step
         })
         .catch(err => {
           setCreateNewEvent(false);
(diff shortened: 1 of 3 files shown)


frontend/add-leads/add-event-step.component.tsx
```diff
@@ -1,7 +1,229 @@
...
-  return <div className="card">Add an event!</div>;
+  const scope = useCss(css);
+  const [isExistingEvent, setIsExistingEvent] = React.useState(true);
+  const [eventDate, setEventDate] = React.useState(
+    dayjs().format("YYYY-MM-DD")
+  );
+  const [allEvents, setAllEvents] = React.useState([]);
+  const [eventName, setEventName] = React.useState("");
+  const [eventLocation, setEventLocation] = React.useState("");
+  const [totalAttendance, setTotalAttendance] = React.useState(1);
```
MeghanBomberger: Could total attendance ever be 0?

joeldenning: I don't think it makes sense to add an event that didn't have any attendance. Especially in the context of creating a lead that is associated with that event. If we have a lead that attended an event, there's at least one attendee

PR #384, opened 2019-11-10
Reuse existing event

joeldenning:
I decided to make the create event and create leads steps their own routes. This resulted in some refactoring. I also finished implementing choosing an existing event (in addition to the already-implemented create new event)


frontend/add-leads/add-event-step.component.tsx
```diff
@@ -106,12 +109,28 @@ export default function AddEventStep(props: AddEventStepProps) {
...
+            {allEvents.map(event => (
+              <option key={event.id} value={event.id}>
+                {event.eventName} ({event.eventDate})
+              </option>
+            ))}
+          </select>
+        </div>
+        <div className="actions">
+          <button className="primary" type="submit" disabled={!existingEventId}>
+            Reuse event
```
MeghanBomberger: Just a wording suggestion. Instead of "Reuse event" would "Add to existing event" convey it better on their end?

PR #388, opened 2019-11-13
Updating get event api to follow API contract.

joeldenning:


backend/apis/events/get-event.api.js
```diff
@@ -26,9 +26,7 @@ app.get("/api/events/:id", (req, res, next) => {
     }
 
     if (event) {
-      res.send({
```
joeldenning: The API for GET /api/events/:id does not specify a json response with an event property at the top level.

Instead of
```json
{
  "events": {
    "id": 1
  }
}
```

it's just
```json
{
  "id": 1
}
```

backend/apis/events/get-event.api.js
```diff
@@ -70,12 +68,12 @@ function getEventById(eventId, cbk, connection) {
       return cbk(err, data, fields);
     }
 
-    const bigEventObj = data[0];
```
joeldenning: The `data` array doesn't always have at least one object in it (when you specify an invalid event id). This was throwing an error in that case.

backend/apis/events/get-event.api.js
```diff
@@ -70,12 +68,12 @@ function getEventById(eventId, cbk, connection) {
...
-
-    if (bigEventObj.length === 0) {
+    if (data.length === 0) {
       return cbk(err, null);
     }
 
+    const bigEventObj = data[0];
```
dvnrsn: error will persist here?

PR #390, opened 2019-11-14
doc - new api to get sum of interaction hours

mexcelus:
Response for API call. Please review request url🙏

### DIFF
--- a/api-docs/get-interactions-hours.md
+++ b/api-docs/get-interactions-hours.md
@@ -1,0 +1,60 @@
+# Number of Interactions per Hour Range.
+
+Returns the total number of clients who have had at least that many client interaction hours in the given time period, along with the list of clients that match the specified criteria.
+
+## Request
+
+_The request should accept query parameters indicating a time range, number of client interaction hours, and current page._
+
+```http
+GET /api/interactions?start=2019-05-06T0&end=2019-05-08T0&=21600&hours=21600&page=1
+```
+
+Notes:
+
+- Result is limited to 100 rows ordered by lastName and then firstName
+- Values for the query should be URL encoded
+- The "page" query parameter defaults to 1
+
+## Response
+
+```json
+{
+  "totalNumberOfClients": 3,
+  "clients": [
+    {
+      "id": 1,
+      "firstName": "John",
+      "lastName": "Doe",
+      "fullName": "John Doe",
+      "gender": "male",
+      "timestamp": "2019-05-06T06:00:00.000Z",
+      "totalDuration": "06:00:00"
+    },
+    {
+      "id": 2,
+      "firstName": "Jane",
+      "lastName": "Doe",
+      "fullName": "Jane Doe",
+      "gender": "female",
+      "timestamp": "2019-05-07T06:00:00.000Z",
+      "totalDuration": "06:15:00"
+    },
+    {
+      "id": 3,
+      "firstName": "Jill",
+      "lastName": "Doe",
+      "fullName": "Jill Doe",
+      "gender": "female",
+      "timestamp": "2019-05-08T06:00:00.000Z",
+      "totalDuration": "06:30:00"
+    }
+  ],
+  "pagination": {
+    "numClients": 3,
+    "currentPage": 1,
+    "pageSize": 100,
+    "numPages": 1
+  }
+}
+```
--- a/api-docs/get-interactions-hours.md
+++ b/api-docs/get-interactions-hours.md
@@ -7,20 +7,23 @@ Returns the total number of clients who have had at least that many client inter
 _The request should accept query parameters indicating a time range, number of client interaction hours, and current page._
 
 ```http
-GET /api/interactions?start=2019-05-06T0&end=2019-05-08T0&=21600&hours=21600&page=1
+GET /api/interactions?start=2019-05-06T0&end=2019-05-08T0&minInteractionMinutes=21600&page=1
 ```
 
 Notes:
 
 - Result is limited to 100 rows ordered by lastName and then firstName
 - Values for the query should be URL encoded
-- The "page" query parameter defaults to 1
+- The 'page' query parameter defaults to 1
+- The 'start' parameter denotes the lower class limit (date and time)
+- The 'end' parameter denotes the upper class limit (date and time)
+- The 'minInteractionSeconds' parameter specifies the least number of hours the query will filter
+- reportDefaultParameters explains values for parameters left blank or empty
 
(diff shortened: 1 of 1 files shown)


api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,60 @@
+# Number of Interactions per Hour Range.
+
+Returns the total number of clients who have had at least that many client interaction hours in the given time period, along with the list of clients that match the specified criteria.
+
+## Request
+
+_The request should accept query parameters indicating a time range, number of client interaction hours, and current page._
+
+```http
+GET /api/interactions?start=2019-05-06T0&end=2019-05-08T0&=21600&hours=21600&page=1
```
joeldenning: Lets put all reports under `/api/reports`. And maybe try to put more into the URL's name so that it's easy to understand why it's returning a list of clients instead of a list of interactions.

```suggestion
GET /api/reports/interaction-hours-by-client?start=2019-05-06T0&end=2019-05-08T0&minInteractionHours=21600&page=1
```

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+Notes:
```
joeldenning: Let's document each of the query params here - `start`, `end`, `minInteractionHours`, and `page`. You already have page documented, but let's document the rest

api-docs/get-interactions-hours.md
joeldenning: Is the `T0` at the end of the string to indicate timezone? If so, that's a nice little touch so that we aren't just assuming UTC in the db query when we likely want MST.

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,60 @@
...
+{
+  "totalNumberOfClients": 3,
+  "clients": [
+    {
+      "id": 1,
+      "firstName": "John",
+      "lastName": "Doe",
+      "fullName": "John Doe",
+      "gender": "male",
+      "timestamp": "2019-05-06T06:00:00.000Z",
```
joeldenning: What is this timestamp? Last time the client was updated? Created? Last interaction created? Last interaction updated?

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,60 @@
...
+```json
+{
+  "totalNumberOfClients": 3,
+  "clients": [
+    {
+      "id": 1,
+      "firstName": "John",
+      "lastName": "Doe",
+      "fullName": "John Doe",
+      "gender": "male",
```
joeldenning: Since the clients table also has the birthday column, let's add `birthday` in as a field, too?

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Result is limited to 100 rows ordered by lastName and then firstName
+- Values for the query should be URL encoded
+- The "page" query parameter defaults to 1
+
+## Response
+
+```json
+{
+  "totalNumberOfClients": 3,
```
joeldenning: Maybe remove this, since `pagination.numClients` already shows this?

```suggestion
```

api-docs/get-interactions-hours.md
joeldenning: The documentation should explain what the query param means, and also indicate what the default value is and whether it's a required param or not. Generally speaking, query params are optional - so it might be good to make all of the params have defaults.

If we add defaults to all the params, it also might be good to add to the response object some properties that explain what the params defaulted to:

```js
{
  "reportParameters": {
    "start": "2019-11-14" // some default
    ...
  }
}
```

api-docs/get-interactions-hours.md
mexcelus: I was considering using the timestamp of when the user was created. In case they need to correlate interactions with client longevity (I wasn't 100% sure of adding it). We can remove it?

api-docs/get-interactions-hours.md
mexcelus: Yep! It could help to query time periods more precisely.

api-docs/get-interactions-hours.md
MeghanBomberger: I don't think we'll need it for this report, but maybe for other reports. If we do keep it here or add it into other reports we should rename it to specify which timestamp it is, such as `dateOfClientCreation` or something similar.

joeldenning: 👍 agreed with Meghan. My vote is to remove since it's not needed right now. And add later as needed.

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,69 @@
...
+    }
+  ],
+  "pagination": {
+    "numClients": 3,
+    "currentPage": 1,
+    "pageSize": 100,
+    "numPages": 1
+  },
+  "reportDefaultParameters": {
+    "start": "[TODAY]", // it will default to currentday
```
joeldenning: You can't put comments into json 😭. Doing so causes this:

image

```suggestion
    "start": "[TODAY]",
```

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,69 @@
...
+    "end": "[TODAY]", // it will default to currentday
```
joeldenning:
```suggestion
    "end": "[TODAY]",
```

api-docs/get-interactions-hours.md
```diff
@@ -0,0 +1,69 @@
...
+      "totalDuration": "06:30:00"
+    }
+  ],
+  "pagination": {
+    "numClients": 3,
+    "currentPage": 1,
+    "pageSize": 100,
+    "numPages": 1
+  },
+  "reportDefaultParameters": {
```
joeldenning: I'm thinking this would be best if it always returns the current parameters, even if they are not the defaults.

Example cases:

```http
/api/reports?minInteractionSeconds=21000
```

Would return:

```json
{
  "reportParameters": {
    "start": "[TODAY]",
    "end": "[TODAY]",
    "minInteractionSeconds": 21000,
    "page": 1
  }
}
```

In other words, let's rename this so that it's not just the default parameters, but the parameters that were actually used:

```suggestion
  "reportParameters": {
```

api-docs/get-interactions-hours.md
mexcelus: My bad lol. I was in a javascript mindset😆, I'll get those removed right away.

PR #391, opened 2019-11-15
adding a /health endpoint for a healtch check for eb

### LINKED ISSUE #356: Make elastic beanstalk health check work
image

seanbw89:
Resolves #356

### DIFF
--- a/backend/apis/health.api.js
+++ b/backend/apis/health.api.js
@@ -1,0 +1,5 @@
+const { app } = require("../server");
+
+app.get("/health", (req, res) => {
+  res.send("Everything is okay.");
+});
--- a/backend/apis/health.api.js
+++ b/backend/apis/health.api.js
@@ -1,5 +1,5 @@
 const { app } = require("../server");
 
-app.get("/health", (req, res) => {
+app.get("/api/health", (req, res) => {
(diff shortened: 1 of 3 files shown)


seanbw89:
Every time that I try to configure this in AWS the website ends up just returning a 502 HTTP status code.

backend/apis/health.api.js
```diff
@@ -0,0 +1,5 @@
+const { app } = require("../server");
+
+app.get("/health", (req, res) => {
```
joeldenning: I'd actually prefer this to be under `/api/health` if that's possible to configure EB for, since the way our web server works is that everything not under `/api` returns the HTML file.

seanbw89: Okay, we can do it this way. Or we can just pass the health check on the 302 redirect. What do you think is going to be the best option?

joeldenning: My guess is that EB's health check looks for a http 200 status and that 302 is just not okay at all (it doesn't care about response body in that case).

If we can't change EB to look for `/api/health` instead of `/health`, then it's okay for us to merge this in as is. It just means the frontend can never have a route at `/health`

PR #392, opened 2019-11-15
Think this should be commands and not command

seanbw89:
image

https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/ebextensions.html

image

### DIFF
--- a/.ebextensions/03-https-redirect-node.config
+++ b/.ebextensions/03-https-redirect-node.config
@@ -94,4 +94,4 @@ files:
  
 container_commands:
   removeconfig:
-command: "rm -f /tmp/deployment/config/#etc#nginx#conf.d#00_elastic_beanstalk_proxy.conf /etc/nginx/conf.d/00_elastic_beanstalk_proxy.conf"
+commands: "rm -f /tmp/deployment/config/#etc#nginx#conf.d#00_elastic_beanstalk_proxy.conf /etc/nginx/conf.d/00_elastic_beanstalk_proxy.conf"
--- a/.ebextensions/03-https-redirect-node.config
+++ b/.ebextensions/03-https-redirect-node.config
@@ -94,4 +94,4 @@ files:
  
 container_commands:
   removeconfig:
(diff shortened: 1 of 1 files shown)


.ebextensions/03-https-redirect-node.config
```diff
@@ -94,4 +94,4 @@ files:
  
 container_commands:
   removeconfig:
-command: "rm -f /tmp/deployment/config/#etc#nginx#conf.d#00_elastic_beanstalk_proxy.conf /etc/nginx/conf.d/00_elastic_beanstalk_proxy.conf"
\ No newline at end of file
+commands: "rm -f /tmp/deployment/config/#etc#nginx#conf.d#00_elastic_beanstalk_proxy.conf /etc/nginx/conf.d/00_elastic_beanstalk_proxy.conf"
```
joeldenning: I forget why I put this here and what it's doing. Is this related to SSL termination at the load balancer? Something else?

seanbw89: This is just an error that I was seeing when trying to setup the health check. Looked it up I think that I found a solution.

joeldenning: Looks like the top of this file explains what it's trying to do. I am worried that by fixing this that it will turn on http/https redirects that are not currently in place, potentially breaking production.

If we merge this in, we'd need to verify that it doesn't break production. I don't think we have https termination turned on at the load balancer level like lines 20 and 23 describe. Can you verify if that is the case. I'm pretty sure that our node web server runs on https in prod, which means the load balancer is not terminating https. If merging this in starts terminating https in prod, then prod will go down.

joeldenning: Actually I just verified [the main server file](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/server.js) and don't see anything about HTTPS.

This could actually explain the cookie issue too -- our server is running on http, but it's being given HTTPs cookies. Which is why my forked version of cookie-session has to exist - to allow for that weirdness.

PR #394, opened 2019-11-17
lead services update

MeghanBomberger:
part of issue #349


backend/apis/leads/update-lead.api.js
```diff
@@ -83,6 +83,47 @@ app.patch("/api/leads/:id", (req, res, next) => {
       );
     }
 
+    const leadServicesChanged = atLeastOne(req.body, "leadServices");
+
+    if (leadServicesChanged && fullLead.leadServices.length > 0) {
+      queries.push("UPDATE leads SET modifiedBy = ? WHERE id = ?;");
+      queryData.push(userId, leadId);
+
+      const oldLeadServices = [...oldLead.leadServices];
```
joeldenning: There doesn't seem to be a purpose for this array existing. We can just delete this line and then change line 93 to be `oldLead.leadServices.map`

backend/apis/leads/update-lead.api.js
```diff
@@ -83,6 +83,47 @@ app.patch("/api/leads/:id", (req, res, next) => {
...
+      const oldLeadServiceIds = oldLeadServices.map(service => {
+        return service.id;
+      });
+      const fullLeadServiceIds = [...fullLead.leadServices];
```
joeldenning: The `fullLeadServiceIds` array does not need to exist. You can just use `fullLead.leadServices`

backend/apis/leads/update-lead.api.js
```diff
@@ -83,6 +83,47 @@ app.patch("/api/leads/:id", (req, res, next) => {
...
+
+      const newLeadServiceIds = [];
+
+      for (let i = 0; i < fullLeadServiceIds.length; i++) {
```
joeldenning: The way this API is implemented does not allow for interest in a service to ever be deleted. I think that absence from the leadServices array should mean that there is no interest in the service.

Also, your algorithm here with the nested for loops could be simplified by using lodash [difference()](https://lodash.com/docs/4.17.15#difference):

```js
const { difference, flatten } = require('lodash')

// do a query to get all the possible services. Maybe do this up top at the same time you're getting the old lead
pool.query(`SELECT id FROM services`, (err, result) => {
  const allServiceIds = result[0]
  const servicesToDelete = difference(fullLead.leadServices, allServiceIds)
  queries.push(...servicesToInsert.map(s => `DELETE FROM leadServices WHERE leadId = ? AND serviceId = ?`))
  queryData.push(...flatten(servicesToInsert.map(s => ([leadId, s.id]))))

  const servicesToInsert = difference(oldLead.services.map(s => s.id), fullLead.leadServices)
  queries.push(...servicesToInsert.map(s => `INSERT INTO leadServices (leadId, serviceId) VALUES (?, ?);`))
  queryData.push(...flatten(servicesToInsert.map(s => ([leadId, s.id]))))
})

```

frontend/add-client/services.component.tsx
```diff
@@ -82,6 +82,12 @@ const css = `
   margin-top: .8rem;
 }
 
+form {
```
joeldenning: All css rules that use kremling should be scoped with the `&` character. See https://kremling.js.org/concepts/scoped-css.html

```suggestion
& form {
```

frontend/util/services-inputs.component.tsx
```diff
@@ -69,11 +67,31 @@ export default React.forwardRef<ServicesRef, IntakeServicesInputsProps>(
 );
 
 const css = `
-& h1 {
-  font-size: 1.8rem;
-}
+	& h3 {
```
joeldenning: tabs -> spaces throughout this template string. Please configure your IDE to always use spaces. https://stackoverflow.com/questions/29972396/how-to-customize-the-tab-to-space-conversion-factor

frontend/view-edit-client/client-home/client-section.component.tsx
```diff
@@ -75,6 +75,23 @@ function getCss() {
       align-items: center;
     }
 
+    & .client-section-content > form > div > label > span {
```
joeldenning: This CSS rule and all of the other ones you added would be significantly easier to understand and maintain if you put a css class onto the span and then referenced that css class directly.

frontend/view-edit-client/client-home/view-edit-intake-info.component.tsx
```diff
@@ -200,6 +200,8 @@ const css = `
   width: 1.6rem;
   margin-left: .8rem;
 }
+
```
joeldenning: Delete this empty lines that aren't needed

frontend/util/services-inputs.component.tsx
```diff
@@ -69,11 +67,31 @@ export default React.forwardRef<ServicesRef, IntakeServicesInputsProps>(
...
+		display: flex;
+		flex-direction: column;
+		justify-content: flex-start;
+		align-items: flex-start;
+	}
 `;
 
+export type IntakeService = {
+  id: number;
+  serviceName: string;
```
joeldenning: I see you moved this from the other file, but forgot to bring over the `serviceDescription` property. Was that on purpose?

frontend/view-edit-lead/edit-lead-inputs/lead-services-inputs.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+    props.lead.leadServices
+  );
+  const leadServicesRef = React.useRef(null);
+
+  React.useEffect(() => {
+    const abortController = new AbortController();
+
+    easyFetch("/api/services", { signal: abortController.signal }).then(data =>
+      setServices(data.services)
+    );
```
joeldenning: To ensure that the error makes it to the window error handler and a toast is shown:

```suggestion
    ).catch(err => {
      setTimeout(() => {
        throw err
      })
    })
```

frontend/view-edit-lead/edit-lead-inputs/lead-services-inputs.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return () => abortController.abort();
+  }, []);
+
+  return (
+    <form onSubmit={handleSubmit} autoComplete="off" className="edit-form">
+      <label>
```
joeldenning: Why does this `<label>` element exist? It doesn't seem like it would be helpful for accessibility, since there are many inputs inside of it, and even labels inside of it.

frontend/view-edit-lead/lead-home/view-edit-lead-services-info.component.tsx
```diff
@@ -1,36 +1,129 @@
...
-          </tr>
-        </thead>
-        <tbody>
-          {lead.leadServices.map(service => {
-            return (
-              <tr key={service.id}>
-                <td align="left">{service.serviceName}</td>
-                <td align="center">{service.programName}</td>
+    <LeadSection
+      title={editing ? "Edit Serivces of Interest" : "Services of Interest"}
```
joeldenning:
```suggestion
      title={editing ? "Edit Services of Interest" : "Services of Interest"}
```

frontend/util/services-inputs.component.tsx
MeghanBomberger: Yes, we didn't seem to use it anywhere, always the serviceName. Did I miss a spot where we used it?

joeldenning: fair enough

backend/apis/leads/update-lead.api.js
```diff
@@ -83,6 +84,47 @@ app.patch("/api/leads/:id", (req, res, next) => {
...
+        for (let i = 0; i < newLeadServiceIds.length; i++) {
+          const serviceId = newLeadServiceIds[i];
+          queries.push(
+            "INSERT INTO leadServices (leadId, serviceId) VALUES (?, ?);"
+          );
+          queryData.push(leadId, serviceId);
+        }
+      }
+
+      const removeLeadSericeIds = _.difference(
```
joeldenning:
```suggestion
      const removeLeadServiceIds = _.difference(
```

frontend/view-edit-client/client-home/client-section.component.tsx
```diff
@@ -75,16 +75,16 @@ function getCss() {
       align-items: center;
     }
 
-    & .client-section-content > form > div > label > span {
+    & .intake-span {
```
joeldenning: Now that I see that these are specific to the intake services, why weren't they put directly into the intake-date-input and view-edit-intake-info files?

It is much much preferable to put your css into the same file and as the component that uses the CSS

frontend/view-edit-client/client-home/client-section.component.tsx
```diff
@@ -75,6 +75,23 @@ function getCss() {
       align-items: center;
     }
 
+    & .intake-span {
```
joeldenning: Now that I see that these are specific to the intake services, why weren't they put directly into the intake-date-input and view-edit-intake-info files?

It is much much preferable to put your css into the same file and as the component that uses the CSS. You prefixing the css class names with `intake-services` is a good indication that this should not be inside of the ClientSection component

frontend/add-client/form-inputs/client-source-inputs.component.tsx
```diff
@@ -55,6 +61,12 @@ export default React.forwardRef<ClientSourceRef, ClientSourceInputsProps>(
   }
 );
 
+const css = `
+& .intake - span {
```
joeldenning:
```suggestion
& .intake-span {
```

PR #401, opened 2019-11-20
Fixing bugs with add leads api

joeldenning:


backend/apis/leads/add-leads.api.js
```diff
@@ -10,7 +10,7 @@ const {
 } = require("../utils/validation-utils");
 
 app.post("/api/leads", (req, res) => {
-  const leads = req.body.data;
```
joeldenning: The api documentation does not call for a `data` property to be in the request body

backend/apis/leads/add-leads.api.js
```diff
@@ -96,9 +98,12 @@ app.post("/api/leads", (req, res) => {
           leadDataArray.push(eventId);
         }
       }
+
+      leadEventsQuery += ";";
     }
 
-    const newLeadQuery = `${addToLeadQuery}; ${setLeadId}; ${leadServicesQuery}; ${leadEventsQuery};`;
```
joeldenning: This query didn't work because it would put semicolons in for all the queries, even if the queries were empty. Mysql doesn't like it when it sees two semicolons right next to each other.

Specifically this would throw an error when the lead services array was passed in as empty.

backend/apis/leads/add-leads.api.js
```diff
@@ -109,7 +114,7 @@ app.post("/api/leads", (req, res) => {
       return databaseError(req, res, err);
     }
 
-    const leadIds = result.map(data => data.insertId).joint(", ");
```
joeldenning: `joint` is not a function -- this API returned a 500 no matter what because of that.

Issue #402, opened 2019-11-20
List Leads API doesn't return leads who don't have interest in any services


joeldenning:
When a lead does not have interest in any services, that lead will never show up in the leads list. This is related to JSON_ARRAYAGG and JSON_OBJECT usage within the list-leads api

PR #403, opened 2019-11-22
resolves issue #402

MeghanBomberger:
For issue #402 

Leads with no services return on list now.
Fixed up UI for the services column of the leads on the front end.
Resolved an additional bug that popped up in the update leads while testing for the list leads.
Fixed up UI for view lead services to show a message if there are no services.


frontend/leads-list/table/lead-services-cell.component.tsx
```diff
@@ -0,0 +1,204 @@
...
+        </div>
+      ) : (
+        <p>No services selected</p>
+      )}
+    </>
+  );
+}
+
+const css = `
+	& .programs-cell {
```
kthompson55: Non-blocking: I looked at the [navbars component](https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/fe1576b894c3a51ee4e483b629b0c43fe5864c6c/frontend/navbar/navbars.component.tsx), and it looks like the indentation of the css object should be two spaces instead of 6.
Could you make sure the formatting of this string is the same as the rest of the repository?

backend/apis/leads/list-leads.api.js
```diff
@@ -152,71 +147,130 @@ app.get("/api/leads", (req, res, next) => {
...
-        lastUpdatedBy: {
-          userId: result.modifiedBy,
-          firstName: result.modifiedByFirstName,
-          lastName: result.modifiedByLastName,
+    const leads = [];
+
+    for (let i = 0; i < leadRows.length; i++) {
+      const getMoreData = mysql.format(
+        `
+        SELECT
```
joeldenning: Two database trips instead of one is not ideal for performance. Nothing is immediately coming to mind as a solution to avoid the two trips, though. Thoughts on how?

backend/apis/leads/list-leads.api.js
```diff
@@ -152,71 +147,130 @@ app.get("/api/leads", (req, res, next) => {
...
+      pool.query(getMoreData, (err, results, fields) => {
+        if (err) {
+          return databaseError(req, res, err);
+        }
+
+        const leadServices = [];
+        const leadEvents = [];
+
+        if (results[0].length > 0) {
+          for (let j = 0; j < results[0].length; j++) {
```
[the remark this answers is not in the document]
joeldenning: Or even better than forEach, just a `.map()` to create the array:

```js
const leadServices = results[0].map(row => ({
  id: row.id,
  serviceName: row.serviceName,
})
```

backend/apis/leads/list-leads.api.js
```diff
@@ -152,71 +147,130 @@ app.get("/api/leads", (req, res, next) => {
...
+            const leadService = {
+              id: results[0][j].serviceId,
+              servicename: results[0][j].serviceName,
```
joeldenning:
```suggestion
              serviceName: results[0][j].serviceName,
```

backend/apis/leads/list-leads.api.js
```diff
@@ -152,71 +147,130 @@ app.get("/api/leads", (req, res, next) => {
...
+              leadRows[i].modifiedByFirstName,
+              leadRows[i].modifiedByLastName
+            ),
+            timestamp: leadRows[i].dateModified
+          }
+        };
+
+        leads.push(lead);
+
+        if (leads.length === leadRows.length) {
```
joeldenning: When is this ever not true.

backend/apis/leads/update-lead.api.js
```diff
@@ -86,7 +86,7 @@ app.patch("/api/leads/:id", (req, res, next) => {
 
     const leadServicesChanged = atLeastOne(req.body, "leadServices");
 
-    if (leadServicesChanged && fullLead.leadServices.length > 0) {
+    if (leadServicesChanged || fullLead.leadServices.length > 0) {
```
joeldenning: This is a duplicate of https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/397.

You messed up the branches when creating PRs - this is something to continue practicing.

backend/apis/leads/list-leads.api.js
MeghanBomberger: leads.length won't equal leadRows.length until it has ittereated over the whole array. I kept running into it res.send ing before the data was ready from the for loop. Will using forEach solve that or do I still need this?

backend/apis/leads/update-lead.api.js
MeghanBomberger: Luckily this one wasn't a git oops. I re-corrected the error while working on this. Forgot I'd already fixed it in the other pending PR. Should I switch it back still?

backend/apis/leads/list-leads.api.js
joeldenning: I just realized that this isn't two database trips, it is `n + 1` database trips (one trip to get the list of leads and one more **per lead**.

This is an absolute showstopper for this PR. We should never be making 101 trips to the database in one API call in order to return 100 leads.

backend/apis/leads/list-leads.api.js
joeldenning: I didn't realize this is in a forEach. This is not how we should be doing things. We should get the data in one step and then send it in the next.

joeldenning: I didn't realize this is in a forEach. This is not how we should be doing things. We should get the data in one step and then send it in the next.

backend/apis/leads/list-leads.api.js
joeldenning: This is still a performance dealbreaker that needs to be fixed. Additionally, there is a race condition here - the code here assumes that the last request to the database is the last response from the database which is not always true.

joeldenning: Oh wait I misread the code. Looks like you have fixed those things. Will take a closer look at this later.

Issue #313, opened 2019-10-01
convert lead to client front end


MeghanBomberger:
Requirements:
* must autofill the create clients data that is provided at the leads stage
* must update the lead to reflect that it was converted

Question for next meeting:
* Is the date converted to client needed data for reports? If so do we need to store that data in the leads database along with the clientId of the client file created?

PR #409, opened 2019-12-09
convert lead to client

MeghanBomberger:
for issue #313


joeldenning:
This PR approaches things in a way that will be hard to maintain - we should pass in initial client state instead of passing in the lead object.

I am closing this and will redo it.

Issue #582, opened 2020-07-15
Add date range for english level, country, client source, age and gender reports


joeldenning:
The date represents the date of intake

joeldenning:
Here is where you can see those reports. We should add date range inputs to them.

http://localhost:8080/reports/english-levels

http://localhost:8080/reports/countries-of-origin

http://localhost:8080/reports/client-sources

http://localhost:8080/reports/ages-and-genders

Here is an example of a report that does a date range: http://localhost:8080/reports/interaction-hours-by-client?maxInclusive=false&minInclusive=false

HappyViki:
@joeldenning `ages-and-genders` has a clients and leads list. From what I understand only clients have date of intake. I'm not sure how to display results. I feel it might be confusing if clients are being filtered based on certain date and leads are not.

I think one solution to make it less confusing is to pick leads from a certain date range as well, except it would be date of sign up.

What do you think?

joeldenning:
Good question - for leads we can use the `dateOfSignUp` column:

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/652e9b479dbb345e92d25f912fef6cd299c282e4/migrations/sqls/20191002015049-leads-and-events-up.sql#L17

joeldenning:
This is done

PR #626, opened 2020-09-06
Add date range on dateOfIntake for english-levels

HappyViki:

Add start and end form fields for [english-levels](http://localhost:8080/reports/english-levels) that grabs rows between two dates (inclusive) based on `dateOfIntake`.

### DIFF
--- a/backend/apis/reports/english-level.api.js
+++ b/backend/apis/reports/english-level.api.js
@@ -18,7 +18,8 @@ app.get(`/api/reports/english-levels`, (req, res) => {
 
   const sql = mysql.format(
     `
-      SELECT COUNT(*) total, demographics.englishProficiency        FROM
+      SELECT COUNT(*) total, demographics.englishProficiency
+      FROM
         (
           SELECT MAX(dateAdded) latestDateAdded, clientId FROM demographics GROUP BY clientId
         ) latestDems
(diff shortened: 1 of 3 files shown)


backend/apis/reports/english-level.api.js
```diff
@@ -18,12 +25,17 @@ app.get(`/api/reports/english-levels`, (req, res) => {
         ) latestDems
         JOIN demographics ON latestDems.latestDateAdded = demographics.dateAdded
         JOIN clients ON clients.id = demographics.clientId
+        JOIN
+        (
+          SELECT dateOfIntake, clientId FROM intakeData
```
joeldenning, line 30: the intakeData table has multiple rows for the same client, so we should only join on the latest rows for each client. See https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/db3866d33f3c5ad0249f362da2f6ae7b4843efa3/backend/apis/reports/service-interest-report.api.js#L38 for more info

joeldenning (commented):
Thanks for your work on this - one thing to update

backend/apis/reports/english-level.api.js
HappyViki: Thanks for pointing that out!

HappyViki:
@joeldenning I added your suggested changes and also I added a table on the results page that shows the params used.

joeldenning (approved):
thanks!

PR #631, opened 2020-09-12
Add date range on dateOfIntake for client-sources

HappyViki:

Add start and end form fields for [client-sources](http://localhost:8080/reports/client-sources) that grabs rows between two dates (inclusive) based on `dateOfIntake`.


joeldenning (approved):
Thank you

PR #632, opened 2020-09-17
Add date range for ages-and-genders

HappyViki:

Add start and end form fields for [ages-and-genders](http://localhost:8080/reports/ages-and-genders) that grabs rows between two dates (inclusive) based on `dateOfIntake` for Clients and `dateOfSignUp` for Leads.

@joeldenning I think it is important to note that the start and end dates mean different things for clients and leads. Let me know if you want me to change how it displays the explanation.
image


joeldenning (approved):
Looks good, thank you.

Issue #643, opened 2020-10-05
Allow for multiple clients on invoices and payments.


joeldenning:
![image](https://user-images.githubusercontent.com/5524384/95140648-9eda6380-072c-11eb-9b8a-2b77ba346db4.png)

![image](https://user-images.githubusercontent.com/5524384/95140665-a7cb3500-072c-11eb-91d9-cf1224aa7eda.png)

When creating and editing invoices and payments, it doesn't allow you to add multiple clients. This was something that Vero asked for when we created the invoice feature.

joeldenning:
Documentation: https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/invoices.md

joeldenning:
Here's the wip branch https://github.com/JustUtahCoders/comunidades-unidas-internal/tree/multiple-clients

PR #654, opened 2020-10-15
Multiple clients for INVOICES

asekito:
closes: #643 

User will now be able to add/remove multiple clients when creating an invoice and have those be reflected in the invoice PDFs.

prgif

Screen Shot 2020-10-15 at 3 34 19 PM

Screen Shot 2020-10-15 at 3 37 22 PM


frontend/view-edit-client/invoices/edit-invoice.component.tsx
```diff
@@ -294,10 +295,14 @@ const EditInvoice = React.forwardRef(function (props: EditInvoiceProps, ref) {
...
       .map((li) => ({ ...li, rate: Number(li.rate) }));
+    const clientIds = clientRef.current
+      .getClients()
+      // @ts-ignore
+      .map((item) => item.clientId);
     const result = {
       ...modifiedInvoice,
       lineItems,
-      clients: props.client ? [props.client.id] : [],
+      clients: props.client ? clientIds : [],
```
joeldenning, line 305: Now that we're guaranteed to have an array of clientIds, I think there's no need to sometimes pass the empty array:

```suggestion
      clients: clientIds,
```

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <div
+      className={maybe(
+        props.containerClass || "",
+        Boolean(props.containerClass)
+      )}
+    >
+      {clients.map((client, idx) => (
+        <div
+          className={maybe(props.clientClass || "", Boolean(props.clientClass))}
+          key={idx}
```
joeldenning, line 38: It's better to use `client.id` here as the `key` instead of the array index, since the array index could theoretically change but the client id wouldn't. The `key` prop is allow about handling the case where the order of arrays change, so passing in the index is not really helpful (even though it makes the pesky warning go away 😄)

```suggestion
          key={client.id}
```

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+  React.useImperativeHandle(ref, () => ({
+    getClients() {
+      return clients;
+    },
+  }));
+
+  const removeClientSelect = () => {
+    let newArr = [...clients];
+    newArr.pop();
```
joeldenning, line 24: This looks like it will always remove the last client from the array, instead of the client that is next to the X that was pressed. We should remove the client for which the X was pressed.

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+import CloseIconButton from "./close-icon-button.component";
+
+const MultiClientSelect = React.forwardRef<
+  MultiClientSelectRef,
+  MultiClientSelectProps
+>((props, ref) => {
+  const [clients, setClients] = React.useState(props.initialClients);
+  const [showEmptyRow, setShowEmptyRow] = React.useState(
+    props.initialClients.length === 0
+  );
```
joeldenning, line 15: This is no longer being used and should be removed:

```suggestion
```

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const removeClientSelect = () => {
+    let newArr = [...clients];
+    newArr.pop();
+    setClients(newArr);
+  };
+
+  return (
+    <div
+      className={maybe(
+        props.containerClass || "",
```
joeldenning, line 31: `props.containerClass` isn't being used, so I'd suggest removing the prop entirely

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+        Boolean(props.containerClass)
+      )}
+    >
+      {clients.map((client, idx) => (
+        <div
+          className={maybe(props.clientClass || "", Boolean(props.clientClass))}
```
joeldenning, line 37: `clientClass` is not being used by the parent component of MultiClientSelect, so let's remove the prop entirely

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+            clientChanged={(...args) => {
+              let newArr = [...clients];
+              newArr[idx] = args[0];
+              setClients(newArr);
+            }}
+            initialClient={client}
+            required
+          />
+          <div style={{ alignSelf: "center", marginTop: "8%" }}>
+            <CloseIconButton close={removeClientSelect} />
```
joeldenning, line 51: To help with removing the client whose X button was clicked:

```suggestion
            <CloseIconButton close={() => removeClientSelect(client.id)} />
```

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+          </div>
+        </div>
+      ))}
+      <button
```
joeldenning, line 55: For better styling

```suggestion
      <button className="secondary"
```

frontend/client-search/single-client/single-client-search-input.component.tsx
```diff
@@ -33,13 +33,13 @@ const SingleClientSearchInput = React.forwardRef<
         clientId: state.clientId,
       };
     }
-  });
+  }, [state.clientId]);
```
joeldenning, line 36: Very good catch 👍

frontend/util/multi-client-select.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <div
+          className={maybe(props.clientClass || "", Boolean(props.clientClass))}
+          key={idx}
+          style={{ display: "flex" }}
+        >
+          <SingleClientSearchInput
+            clientChanged={(...args) => {
+              let newArr = [...clients];
+              newArr[idx] = args[0];
+              setClients(newArr);
```
joeldenning, line 45: This would be simpler if we just called the variable `newClient` instead of collecting it in an `args` array

```suggestion
            clientChanged={(newClient) => {
              setClients(clients.map(c => c.id === client.id ? newClient : c))
            })
```

backend/apis/clients/get-client.api.js
```diff
@@ -188,3 +188,164 @@ function getClientById(clientId, cbk, connection) {
...
 }
+
+exports.getAllClientsById = getAllClientsById;
+
+function getAllClientsById(clients, cbk, connection) {
+  clients = clients.map((item) => Number(item));
+
+  const getClient = mysql.format(
+    `
+    SELECT
```
joeldenning, line 199: Inlining the SQL into a string is the most common pattern in this project, but I now have switched to creating `.sql` files so we can at least get some syntax highlighting and a bit of autocomplete. Let's switch to that:

For example:

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/apis/invoices/get-client-payments.sql

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/314ef28fc185ea052c1ffe5e16b60607fac4a9ed/backend/apis/invoices/get-client-payments.api.js#L13-L15

backend/apis/clients/get-client.api.js
```diff
@@ -188,3 +188,164 @@ function getClientById(clientId, cbk, connection) {
     cbk(err, client);
   });
 }
+
+exports.getAllClientsById = getAllClientsById;
```
joeldenning, line 192: I'm excited about this new function. Let's make the following changes:

1. Create a new file for it - `get-clients-by-id.api.js`
2. Expose it as an API so the browser can also call it
3. Document the API, maybe by adding a new section to https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/api-docs/get-client.md

joeldenning (commented):
Nice work on this! I'm impressed.

I'll review this in more detail later - here are the comments I had before I had to pause.

frontend/view-edit-client/payments/create-edit/edit-payment-info.component.tsx
```diff
@@ -1,6 +1,9 @@
 import React from "react";
 import css from "./edit-payment-info.css";
 import { CreatePaymentStepProps } from "../create-payment.component";
+import MultiClientSelect, {
```
joeldenning, line 4: Let's leave payments-related changes for a different PR. Here are some git commands you can run to undo the changes on this branch and move them to a different branch.

```ts
git checkout master

# Create a new branch based off of master
git checkout -b multiple-clients-payments

# Get the payment files from the multiple-clients branch into the multiple-clients-payments branch
git checkout multiple-clients -- frontend/view-edit-client/payments/create-edit/edit-payment-info.component.tsx frontend/view-edit-client/payments/create-payment.component.tsx frontend/view-edit-client/payments/edit-payment.component.tsx

# You should now have the payment changes on this new branch
git status
git add .
git commit -m "Beginnings of multiple clients on payments"

# Go back to the invoice multiple-clients branch
git checkout multiple-clients
# undo changes made to payments files
git checkout master -- frontend/view-edit-client/payments/create-edit/edit-payment-info.component.tsx frontend/view-edit-client/payments/create-payment.component.tsx frontend/view-edit-client/payments/edit-payment.component.tsx
# verify it says that those files have changed
git status
git add .
git commit -m "Undoing payment changes"
git push
```

In the future, you should avoid getting into this situation by creating a new branch when starting to work on a new feature. The branch can be based off of your existing branch, or off of master. The best one to practice is always branching off of master, since branching of your existing branch requires more advanced git skill.

asekito: Gotcha, fixed it and followed the commands. Had to do --no-verify for the payments branch since the master doesn't have the multi-clients and single client components in it. Hopefully that can be fixed when this PR is merged to the master

backend/apis/clients/get-clients-by-id.api.js
```diff
@@ -0,0 +1,133 @@
...
+} = require("../utils/transform-utils");
+const path = require("path");
+const fs = require("fs");
+const rawGetSql = fs.readFileSync(
+  path.join(__dirname, "./get-clients-by-id.sql")
+);
+
+app.get("/api/clients-by-id", (req, res) => {
+  const validationErrors = Array.isArray(req.query.clientId)
+    ? checkValid(req.query, validArray(validId("clientId")))
```
joeldenning, line 23: validArray is not being imported, so this will throw an error. See line 9 where it should be imported.

joeldenning (approved):
Nice work on this! It turned out to be a lot harder than it should have - thanks for hanging in there with it. I'm merging now and it'll finish deploying to the live environment in about 10 mins.

PR #656, opened 2020-10-19
Implement admin ui for user permissions. Resolves #651.

### LINKED ISSUE #651: Make permissions part of admin settings

joeldenning:
See #651

### DIFF
--- a/backend/apis/users/patch-user.sql
+++ b/backend/apis/users/patch-user.sql
@@ -1,0 +1,12 @@
+UPDATE users SET accessLevel = ? WHERE id = ?;
+
+INSERT IGNORE INTO
+  userPermissions (userId, permission)
+  SELECT ?, 'immigration' WHERE ?
+;
+
+DELETE FROM
+  userPermissions
+  WHERE
+  userId = ? AND permission = 'immigration' AND ?
+;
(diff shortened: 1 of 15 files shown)


frontend/home/home-card.component.tsx
```diff
@@ -87,4 +87,5 @@ type HomeCardProps = {
   title: string;
   iconUrl: string;
   link: string;
+  alt: string;
```
joeldenning, line 90: I added this prop for accessibility and then added the prop to all places that use the HomeCard component

backend/apis/users/patch-user.sql
```diff
@@ -0,0 +1,12 @@
+UPDATE users SET accessLevel = ? WHERE id = ?;
+
+INSERT IGNORE INTO
+  userPermissions (userId, permission)
+  SELECT ?, 'immigration' WHERE ?
```
joeldenning, line 5: This is a way of doing a conditional insert. We inject `WHERE false` when we don't want to insert

backend/apis/users/patch-user.sql
```diff
@@ -0,0 +1,12 @@
...
+;
+
+DELETE FROM
+  userPermissions
+  WHERE
+  userId = ? AND permission = 'immigration' AND ?
```
joeldenning, line 11: The final `AND ?` is a way of doing a conditional delete. We inject `AND false` when we don't want to delete

joeldenning:
New page:

image

joeldenning:
Edit modal:

image

PR #670, opened 2020-10-26
Multiple clients

asekito:
Fixed bug where SingleClientSearchInput was not able to read null "id" from state.client.id when trying to add client interactions or case notes from the home menu. 
Changed state.client.id to state.clientId.


joeldenning:
Closing in favor of #671 since this branch got messed up

PR #671, opened 2020-10-26
fix of a bug state.client.id was null causing Add Client Interaction and Case Note to crash the application

joeldenning:


joeldenning (commented):
Looks good - nice work Andrea

PR #673, opened 2020-10-27
API post and patch request contract for follow ups

asekito:
Closes: #669 
API documentation for the POST and PATCH request and endpoints for client follow ups.

### DIFF
--- a/api-docs/client-follow-up.md
+++ b/api-docs/client-follow-up.md
@@ -6,7 +6,7 @@ The client follow up documents and updates any calls with the client and documen
 
 **_Notes_**
 
-Fields required
+A client follow up must include an associated service id, title of the interaction, the description of the interaction, and a date to follow up with the client.
 
 ### Request
 
--- a/api-docs/client-follow-up.md
+++ b/api-docs/client-follow-up.md
@@ -1,13 +1,17 @@
-# Client follow up note
+# Client follow ups
 
 The client follow up documents and updates any calls with the client and documents the interaction as well as set up a date to follow up with the client.
 
 ## Creating a follow up note on a client
 
+**_Notes_**
+
+Fields required
+
 ### Request
 
 ```http
-POST /api/clients/:clientId/follow-up
+POST /api/clients/:clientId/follow-ups
 ```
 
 ```json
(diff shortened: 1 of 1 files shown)


api-docs/client-follow-up.md
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+```
+
+## Modify a follow up note
+
+### Request
+
+```http
+PATCH /api/clients/:clientId/follow-up/:followUpId
```
joeldenning, line 54:
```suggestion
PATCH /api/clients/:clientId/follow-ups/:followUpId
```

api-docs/client-follow-up.md
```diff
@@ -0,0 +1,91 @@
+# Client follow up note
+
+The client follow up documents and updates any calls with the client and documents the interaction as well as set up a date to follow up with the client.
+
+## Creating a follow up note on a client
+
+### Request
```
joeldenning, line 7: We should add a `**Notes**` section to the request, that indicates which fields are required or optional.

api-docs/client-follow-up.md
```diff
@@ -0,0 +1,91 @@
+# Client follow up note
```
joeldenning, line 1:
```suggestion
# Client follow ups
```

joeldenning (changes requested):
Nice work - a few small comments.

api-docs/client-follow-up.md
```diff
@@ -0,0 +1,91 @@
...
+## Creating a follow up note on a client
+
+### Request
+
+```http
+POST /api/clients/:clientId/follow-up
+```
+
+```json
+{
```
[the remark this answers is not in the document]
asekito: So it's okay if the **PATCH** request doesn't have the `id` property since it is in the params, right? And should the response json also have the id property for the **POST** request?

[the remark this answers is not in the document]
joeldenning: Ah I put the comment here on the wrong line. The `id` should always be present in the Response body, but it doesn't need to be present in the request body. For POST request body, there's no id yet. And for PATCH, it's in the URL params like you mentioned.

asekito (commented):
On a side note, will the user create another follow-up note when they do follow up with the client? Or will they update this note and add another text field? What if they ended up not following up with them or got no response? Should there be a status of "followed up" "not followed up" etc.?

joeldenning (approved):
Looks good, nice work!

PR #674, opened 2020-10-29
Followups schema

asekito:
Closes: #668 
SQL schema compose up and down files for follow up notes.

### DIFF
--- a/migrations/sqls/20201028204537-followups-up.sql
+++ b/migrations/sqls/20201028204537-followups-up.sql
@@ -1,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS followUps (
+  id INT AUTO_INCREMENT PRIMARY KEY,
+  serviceId INT NOT NULL,
+  title VARCHAR(255) NOT NULL,
+  description MEDIUMTEXT,
+  dateOfContact DATE,
+  appointmentDate DATETIME,
+  addedBy INT NOT NULL,
+  updatedBy INT NOT NULL,
+  FOREIGN KEY (serviceId) REFERENCES services(id),
+  FOREIGN KEY (addedBy) REFERENCES users(id),
+  FOREIGN KEY (updatedBy) REFERENCES users(id)
+)
--- a/migrations/sqls/20201028205336-followups-down.sql
+++ b/migrations/sqls/20201028205336-followups-down.sql
@@ -1,0 +1,1 @@
+DROP TABLE followUps;
--- a/api-docs/client-follow-up.md
+++ b/api-docs/client-follow-up.md
@@ -6,7 +6,7 @@ The client follow up documents and updates any calls with the client and documen
 
 **_Notes_**
 
-Fields required
+A client follow up must include an associated service id, title of the interaction, the description of the interaction, and a date to follow up with the client.
 
 ### Request
 
--- a/api-docs/client-follow-up.md
+++ b/api-docs/client-follow-up.md
@@ -1,13 +1,17 @@
-# Client follow up note
+# Client follow ups
 
 The client follow up documents and updates any calls with the client and documents the interaction as well as set up a date to follow up with the client.
 
 ## Creating a follow up note on a client
 
+**_Notes_**
+
+Fields required
+
 ### Request
 
 ```http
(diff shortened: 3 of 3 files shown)


migrations/sqls/20201028205336-followups-down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE followUps;
```
joeldenning, line 1: In situations where the up migrations fail, it's nice to make the down migrations flexible enough so that they don't also fail.

```suggestion
DROP TABLE IF EXISTS followUps;
```

migrations/sqls/20201028204537-followups-up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS followUps (
```
joeldenning, line 1: This table should have a `clientId` column that is a foreign key to the `clients` table.

migrations/sqls/20201028204537-followups-up.sql
```diff
@@ -0,0 +1,13 @@
...
+  id INT AUTO_INCREMENT PRIMARY KEY,
+  serviceId INT NOT NULL,
+  title VARCHAR(255) NOT NULL,
+  description MEDIUMTEXT,
+  dateOfContact DATE,
```
joeldenning, line 6: Why is dateOfContact a DATE but appointmentDate is DATETIME?

joeldenning (commented):
Looking good, a few comments.

migrations/sqls/20201028204537-followups-up.sql
joeldenning: Also, I think `dateOfContact` should be NOT NULL. Appointment date is optional though

migrations/sqls/20201028204537-followups-up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS followUps (
+  id INT AUTO_INCREMENT PRIMARY KEY,
+  serviceId INT NOT NULL,
```
joeldenning, line 3: As I looked this over, it crossed my mind that maybe a call could discuss multiple CU services? I guess we could have them create two follow ups in that case?

migrations/sqls/20201028204537-followups-up.sql
asekito: Oops, that should have stayed DATE, might have been the autofill/smart guess vscode feature.

migrations/sqls/20201028204537-followups-up.sql
asekito: Multiple CU services would be useful. Would it be incoming as an array of service ids in the json object? Might be something to ask Vicky or Mayra @joeldenning

joeldenning: To implement multiple invoices, a separate table would be best - one that has the follow up id and the service id. A "join table." Maybe called `followUpServices`

migrations/sqls/20201028204537-followups-up.sql
joeldenning: Let's have them both as DATETIME - that way we can capture exact hour of an appointment if needed.

migrations/sqls/20201028204537-followups-up.sql
```diff
@@ -0,0 +1,21 @@
...
+  title VARCHAR(255) NOT NULL,
+  description MEDIUMTEXT,
+  dateOfContact DATETIME,
+  appointmentDate DATETIME,
+  addedBy INT NOT NULL,
+  updatedBy INT NOT NULL,
+  FOREIGN KEY (clientId) REFERENCES clients(id),
+  FOREIGN KEY (addedBy) REFERENCES users(id),
+  FOREIGN KEY (updatedBy) REFERENCES users(id),
+)
```
joeldenning, line 13: I thought MySQL required semicolons. Is this working without them? I would add them in, to be safe.

joeldenning (commented):
The following comment still needs to be addressed.

https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/674#discussion_r514516482

PR #678, opened 2020-11-03
Followup post endpoint

asekito:
Closes: #667 
Posts the follow up information as per the follow up API contract and responds with the information from the posted follow up.

Example response object
[
     RowDataPacket {
         id: 22,
         clientId: 1,
         title: 'fadsf Title',
         description: 'fdsaf Test description',
         dateOfContact: 2020-08-19T07:00:00.000Z,
         appointmentDate: 2020-09-01T07:00:00.000Z,
         serviceIds: '[8, 9]',
         createdBy: '{"userId": 1, "lastName": "User", "firstName": "Local"}',
         lastUpdatedBy: '{"userId": 1, "lastName": "User", "firstName": "Local"}'
     }
 ]


backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -0,0 +1,94 @@
...
+
+  pool.query(insertFollowUpSql, (err, insertResult) => {
+    if (err) {
+      return databaseError(req, res, err);
+    }
+
+    let query = "";
+
+    serviceIds.forEach((id) => {
+      query += `
```
joeldenning, line 65: use `mysql.format()` here to avoid sql injection

backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -0,0 +1,94 @@
...
+        if (err) {
+          return databaseError(err);
+        }
+        res.send(result);
+      });
+    });
+  });
+});
+
+function formattedFollowUpResponse(id, errBack) {
```
joeldenning, line 85: Generally when using JSON_ARRAYAGG and maybe also JSON_OBJECT you have to call `JSON.parse()` on the returned fields, because the mysql driver gives you the data as a string instead of an array or object. Have you confirmed whether that is the case for the `serviceIds`, `createdBy`, and `modifiedBy` properties?

asekito: Ah my bad, I thought that was handled on the front end! Will parse on server end!

asekito (commented):
now responding with object similar to this:
{
  appointmentDate: "2020-09-01T07:00:00.000Z"
  clientId: 1
  createdBy: {userId: 1, lastName: "User", firstName: "Local"}
  dateOfContact: "2020-08-19T07:00:00.000Z"
  description: "fdsaf Test description"
  id: 41
  lastUpdatedBy: {userId: 1, lastName: "User", firstName: "Local"}
  serviceIds: (2) [8, 10]
  title: "fadsf Title"
}

backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -0,0 +1,106 @@
...
+  const query = mysql.format(getFollowUpSql, [id, id]);
+  pool.query(query, (err, followUpResult) => {
+    if (err) {
+      return errBack(err, null);
+    } else {
+      let formmattedResult = { ...followUpResult[0] };
```
joeldenning, line 97:
```suggestion
      let formattedResult = { ...followUpResult[0] };
```

joeldenning (commented):
One last comment - a typo. After that is fixed I'll merge this.

Issue #636, opened 2020-09-22
Add zip code report


joeldenning:
Show all zip codes, and how many clients in each. Also show the city name next to the zip code

PR #689, opened 2020-11-07
first cut of zipcode report, modified files to render new report on w…

oliviaclyde:
…eb app, still working on api to get it to serve to frontend

I have the structure in place. The mySQL query works in the terminal. Still trying to work on the api to serve to the front end results component file, but wanted to get what I had up.


oliviaclyde:
Addresses issue #636

backend/apis/reports/zipcode-report.api.js
```diff
@@ -0,0 +1,46 @@
...
+    return invalidRequest(res, validationErrors);
+  }
+
+  const startDate = req.query.start || "2000-01-01T0";
+  const endDate = req.query.end || "3000-01-01T0";
+
+  const sql = mysql.format(
+    `
+        SELECT zip, city, COUNT(*) as clientCount
+        FROM contactInformation GROUP BY zip ORDER BY clientCount DESC;
```
joeldenning, line 24: The contactInformation table is contains both historical and current rows. Because of this, we have to select only the latest (current) rows from the table.

The following code shows how to do that:

https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/d99947b3edd33a687c56e4bce3719342981574db/backend/apis/clients/get-clients-by-id.sql#L27-L36. In that code, you should copy paste the entire `contactInfo` section (line 27 to 36). The contactInfo temporary table is the one that you can use in this zip code report.

backend/apis/reports/zipcode-report.api.js
```diff
@@ -0,0 +1,46 @@
...
+        `,
+    [startDate, endDate, startDate, endDate]
+  );
+
+  pool.query(sql, (err, result) => {
+    if (err) {
+      return databaseError(req, res, err);
+    }
+
+    const [clientZipcode, clientCity, clientCount] = results;
```
joeldenning, line 34: This code will throw an error, since `results` is not defined. Additionally, the destructuring here is also incorrect - the  `result` variable here will be an array where each row is a zip code. But the code here extracts only the first three rows and treats them as each row representing a different field.

What this suggests is that you haven't executed this code yet on your computer (although I'm sure you've tried to!). Feel free to message me on Slack if you need help debugging the code and ensuring that it's executing.

backend/apis/reports/zipcode-report.api.js
```diff
@@ -0,0 +1,46 @@
+const { app, invalidRequest, pool, databaseError } = require("../../server");
+const { checkValid, nullableValidDate } = require("../utils/validation-utils");
+const mysql = require("mysql");
+const _ = require("lodash");
+const ClientCapability = require("twilio/lib/jwt/ClientCapability");
+
+app.get(`/api/reports/zipcode-report`, (req, res) => {
```
joeldenning, line 7: A couple conventions we use in this project:

1. nouns in API endpoints are plural
2. no need to put in `-report` since already have `/reports` earlier in the URL

```suggestion
app.get(`/api/reports/client-zipcodes`, (req, res) => {
```

frontend/reports/reports.components.tsx
```diff
@@ -80,6 +82,11 @@ export default function Reports(props: ReportsProps) {
               title="Ages and Genders"
             />
             <AgesAndGendersResults path="ages-and-genders/results" />
+            <ClientZipcodeParams
+              path="client-zipcode"
```
joeldenning, line 86: The way that the reports are set up requires that the frontend route match the backend API's path. So this line needs to match https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/689/files#diff-ae17fbe5ac7201c5776b7888b1827149906ec26ddc054a936017aa1f5825886cR7

```suggestion
              path="client-zipcodes"
```

joeldenning (changes requested):
This is a good first pass - you followed the patterns well. I've found a few things to address, but it's not too far off from being mergeable.

backend/apis/reports/zipcode-report.api.js
oliviaclyde: Okay. I modified my sql query with the above code. I tested it in the console and it returns correctly.

backend/apis/reports/zipcode-report.api.js
oliviaclyde: Yes, good catch. The singular vs plural consistency tripped me up.

frontend/reports/reports.components.tsx
oliviaclyde: I have fixed this. Got tripped up with the singular vs plural

backend/apis/reports/zipcode-report.api.js
oliviaclyde: I worked with Ben to get the data to send correctly. Once we got that to work, then he helped me create a map that would map each result as a row in each column.

joeldenning:
here’s where you can see why the build is failing: https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/689/checks?check_run_id=1712335821. The reason is that prettier, the code formatter, was not run on the files. This normally occurs automatically when running git commit, so you must have skipped that step accidentally. To fix, run `pnpm run format` which will fix the code formatting.

backend/apis/reports/zipcode-report.api.js
joeldenning: This line of code can be removed, since none of the variables are being used later on in the file

backend/apis/reports/zipcode-report.api.js
```diff
@@ -0,0 +1,58 @@
...
+  if (validationErrors.length > 0) {
+    return invalidRequest(res, validationErrors);
+  }
+
+  const startDate = req.query.start || "2000-01-01T0";
+  const endDate = req.query.end || "3000-01-01T0";
+
+  const sql = mysql.format(
+    `
+      SELECT zip, city, COUNT(*) as clientCount
```
joeldenning, line 23: By selecting both the zip and city, we will now have multiple rows for the same zip code, in the situations where the zip code doesn't match city boundaries. This might be confusing when reading the report, since the report is a zip code report but would not show you the grand total for a zip code in the situations where there were two cities for the same zip code.

To resolve, we could either show a grand total for each zip code, or we could remove the city from this query so that there's guaranteed to only have one row per zip code. The latter option is simpler to implement and totally fine for CU - they know most of the zip codes and can look up the rest.

frontend/reports/client-zipcode/client-zipcode-results.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+            <th>Zipcode</th>
+            <th>City</th>
+            <th>Client Count</th>
+          </tr>
+        }
+        contentRows={
+          <>
+            {data && data.results.map((result) => 
+              <tr>
+                <td>{result.zip}</td>
```
joeldenning, line 62: We should handle the situation where `result.zip` is null. Maybe like this:

```suggestion
                <td>{result.zip || "(No Zip)"}</td>
```

backend/apis/reports/zipcode-report.api.js
oliviaclyde: Good call out. I have removed the city.

frontend/reports/client-zipcode/client-zipcode-results.component.tsx
oliviaclyde: Thanks for point this out. Best-practice is to cover edge cases. I did double check the frontend, and the form has the zip code as a required field. But as a developer, I can't rely on the frontend to handle edge cases. I need to handle them in the backend.

backend/apis/reports/zipcode-report.api.js
oliviaclyde: Done!

joeldenning:
I've pushed https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/689/commits/3d87ae45a2b53a01a41b09a8804e6abc6bc6726f with a couple small fixes. Will merge once the build passes. Thanks!

PR #690, opened 2020-11-13
Issue 684

### LINKED ISSUE #684: When viewing event, show list of leads from that event.

TheMcMurder:
Fixes #684 

Screen Shot 2020-11-13 at 4 07 18 PM

Notes:
- I didn't implement it but it might be nice to also see which clients were created from the event by looking at this data as well: https://github.com/JustUtahCoders/comunidades-unidas-internal/blob/master/backend/apis/events/get-event.api.js#L83
- I debated restructuring the API response a bit more but went for the simplest possible change

### DIFF
--- a/api-docs/events.md
+++ b/api-docs/events.md
@@ -47,7 +47,21 @@ GET /api/events/:id
     "lastName": "Denning",
     "fullName": "Joel Denning",
     "timestamp": "2019-05-06T06:00:00.000Z"
-  }
+  },
+  "leads": [
+    {
+      "leadId": 1,
+      "firstName": "Justin",
+      "lastName": "McMurdie",
+      "fullName": "Justin McMurdie"
+    },
+    {
+      "leadId": 2,
+      "firstName": "Obi-Wan",
+      "lastName": "Kenobi",
+      "fullName": "Obi-Wan Kenobi"
+    }
+  ]
 }
 ```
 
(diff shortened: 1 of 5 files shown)


api-docs/events.md
```diff
@@ -47,7 +47,21 @@ GET /api/events/:id
     "lastName": "Denning",
     "fullName": "Joel Denning",
     "timestamp": "2019-05-06T06:00:00.000Z"
-  }
+  },
+  "leads": [
+    {
+      "leadId": 1,
```
TheMcMurder, line 53: I opted for `leadId` here instead of just `id`. I'm happy to change it if you don't like it.

backend/apis/events/get-event.api.js
```diff
@@ -62,7 +62,7 @@ function getEventById(eventId, cbk, connection) {
       WHERE events.id = ? AND isDeleted = false;
 
       SELECT
-        leadEvents.leadId, leads.gender, leads.leadStatus
+        leadEvents.leadId, leads.gender, leads.leadStatus, leads.firstName, leads.lastName
```
TheMcMurder, line 65: This API was already getting all the leads for the event. I just needed to add their first and last names.

frontend/view-edit-events/view-event.component.tsx
```diff
@@ -121,6 +121,7 @@ export type SingleEvent = {
   totalConvertedToClients?: number;
   leadGenders?: any;
   clientGenders?: any;
+  leads?: any;
```
TheMcMurder, line 124: Feels like cheating a bit, but it was being done on `leadGenders` and `clientGenders`

joeldenning: fair enough

joeldenning (approved):
Thanks!

Issue #681, opened 2020-11-04
Add duration to follow ups


joeldenning:

asekito:
closed with issue-665 branch

PR #697, opened 2020-11-21
Issue 665

asekito:
Closes: #665 #681 
- Adds follow ups to client logs
- Displays follow up to client history on the front end with color "coral" (change if needed)
- Added duration "00:00:00" to follow ups
- Fixed follow up creation timestamp to current time of SQL addition
Screen Shot 2020-11-21 at 4 00 30 PM

(Still need to fix patch req as per the comment )


backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -72,6 +76,15 @@ app.post("/api/clients/:clientId/follow-ups", (req, res) => {
       );
     });
 
+    query += insertActivityLogQuery({
+      detailId: insertResult.insertId,
+      clientId,
+      title,
```
joeldenning, line 82: The title here should be `Client received follow-up regarding <serviceName>`

backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -72,6 +76,15 @@ app.post("/api/clients/:clientId/follow-ups", (req, res) => {
...
+      description,
```
joeldenning, line 83: Description should be `null`

 ```suggestion
      description: null,
```

File: backend/apis/clients/follow-ups/patch-follow-up.api.js [hunk 3]
```diff
@@ -78,6 +81,9 @@ app.patch("/api/clients/:clientId/follow-ups/:followUpId", (req, res) => {
         [id, newFollowUp.id]
       );
     });
+
+    // update client logs here as well???? TBD
```
joeldenning, line 85: we should insert a new log, and then update the old logs `idOfUpdatedLog` column.

migrations/sqls/20201028204537-followups-up.sql
```diff
@@ -5,6 +5,7 @@ CREATE TABLE IF NOT EXISTS followUps (
   description MEDIUMTEXT,
   dateOfContact DATETIME NOT NULL,
   appointmentDate DATETIME,
+  duration TIME DEFAULT '0:00:00' NOT NULL,
```
joeldenning, line 8: Modifying a migration doesn't work, since the migration has already run on the production database. Instead, create a new migration that does `ALTER TABLE followUps ADD COLUMN`

backend/apis/clients/follow-ups/patch-follow-up.api.js
```diff
@@ -63,31 +66,76 @@ app.patch("/api/clients/:clientId/follow-ups/:followUpId", (req, res) => {
...
+          updateResult[updateResult.length - 1][0]["services"]
+        }`,
+        description: null,
+        logType: "follow-up",
+        addedBy: user.id,
+      });
+
+      clientLogUpdate += mysql.format(
+        `
+        WITH oldLog AS (
```
joeldenning, line 113: With clauses are advanced, nice work.

joeldenning (approved):
Nice work

PR #701, opened 2020-12-02
Fix bug with creating follow ups that had no services.

joeldenning:
Previously, a serviceIds array that was empty would result in a server error. This fixes that.

### DIFF
--- a/backend/apis/clients/follow-ups/create-follow-up.api.js
+++ b/backend/apis/clients/follow-ups/create-follow-up.api.js
@@ -80,8 +80,6 @@ app.post("/api/clients/:clientId/follow-ups", (req, res) => {
       [serviceIds.length > 0 ? serviceIds : null]
     );
 
-    console.log(query);
-
     pool.query(query, (err, joinResult) => {
       if (err) {
         return databaseError(req, res, err);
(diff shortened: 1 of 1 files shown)


backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -32,7 +32,9 @@ app.post("/api/clients/:clientId/follow-ups", (req, res) => {
     ...checkValid(req.body, nullableValidDateTime("appointmenetDate")),
   ];
 
-  const { clientId } = req.params;
+  let { clientId } = req.params;
+
+  clientId = Number(clientId);
```
joeldenning, line 37: URL params always come into express as strings, even though the clientId should be a number. Using it as a string wasn't causing issues, but I fixed this when I noticed it when logging sql queries.

backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -58,10 +60,6 @@ app.post("/api/clients/:clientId/follow-ups", (req, res) => {
     user.id,
   ]);
 
-  const insertFollowUpServicesSql = mysql.format(`
```
joeldenning, line 61: This was dead code

backend/apis/clients/follow-ups/create-follow-up.api.js
```diff
@@ -79,20 +77,26 @@ app.post("/api/clients/:clientId/follow-ups", (req, res) => {
 
     query += mysql.format(
       `SELECT GROUP_CONCAT(services.serviceName SEPARATOR ', ') services FROM services WHERE id IN (?);`,
-      [serviceIds]
+      [serviceIds.length > 0 ? serviceIds : null]
```
joeldenning, line 80: `IN (NULL)` is a way of executing the sql statement and guaranteeing it returns nothing

PR #715, opened 2020-12-23
UI for follow-ups in Interactions by Service report

leahjlou:
I noticed a couple of weird things while adding this UI.

1. Follow-ups aren't required to have an associated service. The follow-ups with unspecified services don't show up anywhere in this report or in the total. It seems misleading/confusing to me. Thoughts on adding an "unspecified" row? (this would also require a backend change, happy to do that)

2. If a follow-up is associated with multiple services, the "grand total" counts that same follow-up twice in the sums for count & duration. It seems like we want to change the backend reporting logic here -- I'm happy to do this, but because it's a nontrivial change I wanted to check if there is any reason we might want to keep this or maybe even change a follow-up to only be associated with 1 service.

For example:

Created a follow up for a client with two different services selected:
Screen Shot 2020-12-22 at 6 25 17 PM

Report shows follow up count/duration correctly on individual line items, but the total counts the same follow-up twice:
Screen Shot 2020-12-22 at 6 25 54 PM

### DIFF
--- a/backend/apis/reports/interactions-by-service.api.js
+++ b/backend/apis/reports/interactions-by-service.api.js
@@ -305,7 +305,7 @@ app.get(`/api/reports/interactions-by-service`, (req, res) => {
 
     const grandTotal = {
       numInteractions: _.sum(programTotals.map((p) => p.numInteractions)),
-      numFollowUps: _.sum(followUpProgramTotals.map((p) => p.numInteractions)),
+      numFollowUps: _.sum(followUpProgramTotals.map((p) => p.numFollowUps)),
       numClients: totalClients[0].numClients,
       totalInteractionSeconds: _.sum(
         programTotals.map((p) => p.totalInteractionSeconds)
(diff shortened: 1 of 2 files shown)


joeldenning:
> Thoughts on adding an "unspecified" row?

Yes let's do that.

> The "grand total" counts that same follow-up twice

Yes this is a problem that we should fix. Associating a follow up with multiple services is as-designed. Not sure the best way to do this. Maybe just the grand total row is corrected?

leahjlou:
@joeldenning 

> Associating a follow up with multiple services is as-designed. Not sure the best way to do this. Maybe just the grand total row is corrected?

Yeah, I think just correcting the grand totals to not double-count the followups would be good. I'll add that in on this branch (and the unspecified row). Thanks for your input!

leahjlou:
Added an unspecified row for follow-ups that aren't associated with a program/service, and corrected the total row to display the real total (not double-counting a single follow-up).

It is working correctly -- the end experience feels a little confusing to me, but it's probably OK. In these screenshots, you can see that the totals row is correctly summing the follow-ups, but it might seem weird that it doesn't actually display the total if you sum the columns above, because in this case, one of the follow-ups is associated with two different services, and two of the follow-ups are unspecified. I think it's okay, but wanted to run it by you. I can't think of a way to display the data that is more accurate than this.

Screen Shot 2020-12-29 at 12 10 00 PM
Screen Shot 2020-12-29 at 12 11 43 PM

joeldenning (approved):
Thanks

PR #769, opened 2021-03-04
Issue 742

nick-nack-attack:
• Changed `top: 45;` to `top: 45rem;`
• Added `.idea` to `.gitignore` for Webstorm IDE

### DIFF
--- a/frontend/leads-list/table/desktop-leads-table.component.tsx
+++ b/frontend/leads-list/table/desktop-leads-table.component.tsx
@@ -324,7 +324,7 @@ const css = `
   }
 
   & .leads-table.advanced-search th {
-    top: 47.5;
+    top: 45rem;
   }
 
   & .leads-table th button {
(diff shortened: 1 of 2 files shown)


frontend/leads-list/table/desktop-leads-table.component.tsx
```diff
@@ -324,7 +324,7 @@ const css = `
   }
 
   & .leads-table.advanced-search th {
-    top: 47.5;
+    top: 45rem;
```
joeldenning, line 327: In my comment on Github, I think I made a mistake. I did not mean to change from `47.5` to `45` - just meant to add the `rem`. Were you able to confirm whether 47.5 or 45 is the correct value for scroll when advanced search is expanded?

joeldenning: I confirmed this myself today. Looks good.

joeldenning (approved):

PR #770, opened 2021-03-04
Issue 768

nick-nack-attack:
• The client wants NAME, NUMBER, and ADDRESS to be exported but I don't have that data on the frontend.
• Also, the data needs to be reformatted to let `exportToCsv` export a properly formatted csv file (i.e. an array of objects).
• I hard-coded test data to the top of `bulk-export-modal.component.tsx` to demonstrate the frontend is working.
• @joeldenning I'm open to what direction you'd like it to go from here.


.idea/comunidades-unidas-internal.iml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
joeldenning, line 1: This file should not be committed - please remove along with the other files.

frontend/util/export-to-csv.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+              }
+              return cell;
+            })
+            .join(separator);
+        })
+        .join("\n");
+
+    const blob = new Blob([csvContent], { type: "text/csv;charset=utf-8;" });
+    if (navigator.msSaveBlob) {
+      // IE 10+
```
joeldenning, line 33: We don't need to support IE10 or IE11, so this if statement could be removed.

frontend/util/export-to-csv.component.tsx
```diff
@@ -0,0 +1,57 @@
+// from https://codeburst.io/export-objects-array-as-csv-using-typescript-643bf4f794d9
+
+export class CsvDataService {
```
joeldenning, line 3: Using classes for functions like this in javascript is unnecessary and unconventional in the react community. Better is just to export the functions directly:

```js
export function exportToCsv(filename, rows) {
```

frontend/util/export-to-csv.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+      keys.join(separator) +
+      "\n" +
+      rows
+        .map((row) => {
+          return keys
+            .map((k) => {
+              let cell = row[k] === null || row[k] === undefined ? "" : row[k];
+              cell =
+                cell instanceof Date
+                  ? cell.toLocaleString()
```
joeldenning, line 20: Let's use dayjs instead of `toLocaleString()`.

```js
dayjs(cell).format('YYYY-MM-DD HH:mm:ss')
```

frontend/bulk-export/bulk-export-modal.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+  },
+  {
+    firstName: "Jack",
+    lastName: "Johnson",
+    phone: "9203929303",
+    address: "147 Main street, Salt Lake City, UT 32231",
+  },
+];
+
+export default function BulkExportModal(props: BulkExportModalProps) {
```
joeldenning, line 27: This modal will need to handle when all clients are selected across all pages. Currently only one page of clients is loaded at a time, but for exporting all at once you'd need to go through and download each page.

joeldenning (commented):
This is a good start, thanks for your contribution.

joeldenning:
Since the comments weren't addressed, I created a separate PR for this at https://github.com/JustUtahCoders/comunidades-unidas-internal/pull/776. I'm closing this one in favor of #776

PR #779, opened 2021-03-23
Get services api 756

zaratejavier:


backend/apis/services/list-services.api.js
```diff
@@ -18,8 +19,11 @@ app.get("/api/services", (req, res, next) => {
   const includeInactive = req.query.includeInactive === "true";
 
   const getServices = mysql.format(`
-    SELECT *, services.id AS serviceId, programs.id AS programId
-    FROM services JOIN programs WHERE services.programId = programs.id;
+    SELECT * FROM services WHERE isActive = true;
```
joeldenning, line 22: I forgot the `isActive` check is done later on (see line 63). Because of that, we can remove the check here.

```suggestion
    SELECT * FROM services;
```

joeldenning (approved):

PR #786, opened 2021-04-09
resolved #759. update client interaction API changes

### LINKED ISSUE #759: Update client interaction API changes

zaratejavier:


backend/apis/clients/client-interactions/patch-client-interaction.api.js
```diff
@@ -189,6 +203,26 @@ app.patch("/api/clients/:clientId/interactions/:interactionId", (req, res) => {
           );
         }
 
+        req.body.customQuestions.map((question) => {
+          queries.push(
+            mysql.format(
+              `
+                UPDATE clientInteractionCustomAnswers
```
joeldenning, line 210: This is incorrect and won't work under several circumstances. Specifically, when the request body provides an answer to a new question, this will fail.

To fix, we should delete all the rows from clientinteractionCustomAnswers for this clientInteraction, and then re-insert. I was wrong when we discussed this earlier and I said not to do that - I thought that we were inserting a new clientInteraction, but we aren't.

backend/apis/clients/client-interactions/patch-client-interaction.api.js
```diff
@@ -189,6 +203,33 @@ app.patch("/api/clients/:clientId/interactions/:interactionId", (req, res) => {
...
+            `
+            DELETE from clientInteractionCustomAnswers where interactionId = ?;
+            `,
+            [req.params.interactionId]
+          );
+          queries.push(deleteQuery);
+          const createCustomAnswerQueries = req.body.customQuestions.map(
+            (question) =>
+              mysql.format(
+                mysql.format(`
```
joeldenning, line 217: Why the two mysql.format() calls inside of each other? We should only need one.

joeldenning (approved):

PR #800, opened 2021-04-22
Issue 789-3

zaratejavier:

PR #803, opened 2021-04-26
Issue 789-4

zaratejavier:
image
image

### DIFF
--- a/frontend/reports/client-zipcode/client-zipcode-results.component.tsx
+++ b/frontend/reports/client-zipcode/client-zipcode-results.component.tsx
@@ -102,7 +102,7 @@ export default function ClientZipcodeResults(props) {
 
   function getCsvOptions(): Promise<CsvOptions> {
     const countyZips = [];
-    Object.keys(data.zipsByCounty).map((county) => {
+    Object.keys(data.zipsByCounty).forEach((county) => {
       const zips = data.zipsByCounty[county];
 
       countyZips.push({
--- a/frontend/reports/client-zipcode/client-zipcode-results.component.tsx
+++ b/frontend/reports/client-zipcode/client-zipcode-results.component.tsx
@@ -6,6 +6,7 @@ import CollapsibleTableRows, {
   ToggleCollapseButton,
 } from "../shared/collapsible-table-rows.component";
 import { sumBy } from "lodash-es";
+import { CsvOptions } from "../../util/csv-utils";
 
 export default function ClientZipcodeResults(props) {
   const { isLoading, data, error } = useReportsApi(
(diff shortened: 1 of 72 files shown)


frontend/reports/poverty-line/poverty-line-results.component.tsx
```diff
@@ -3,6 +3,7 @@ import { useReportsApi } from "../shared/use-reports-api";
 import BasicTableReport from "../shared/basic-table-report.component";
 import { formatPercentage } from "../shared/report.helpers";
 import { sum, values } from "lodash-es";
+import { CsvOptions } from "../../util/csv-utils";
```
joeldenning, line 6: It seems you may have mixed branches here - this code was in #800 but is also in this pull request. Ideally each pull request would be independent and not have the same code in them.

There are some ways to fix it with git, but the easiest might be to address the feedback I left in #800 and get that merged. After it's merged, you merge master into issue-789-4 and this pull request will only have the two new reports rather than the two reports from #800

frontend/reports/service-interests/service-interests-results.component.tsx
```diff
@@ -70,4 +72,36 @@ export default function ServiceInterestsResults(props) {
       footerRows={null}
     />
   );
+
+  function getCsvOptions(): Promise<CsvOptions> {
+    const interestResults = [];
+    data.programs.map((program) => {
```
joeldenning, line 78: When you're not storing the result of `.map()`, there's no point in using `.map()`. Instead, you should use `.forEach()`.

```suggestion
    data.programs.forEach((program) => {
```

frontend/reports/service-interests/service-interests-results.component.tsx
```diff
@@ -70,4 +72,36 @@ export default function ServiceInterestsResults(props) {
...
+      interestResults.push({
+        Program: program.programName,
+        Service: "All",
+        "Clients interested": program.clientsInterested.toLocaleString(),
+        "Leads interested": program.leadsInterested.toLocaleString(),
+      });
+
+      groupedServices[program.programId].map((service) => {
```
joeldenning, line 86:
```suggestion
      groupedServices[program.programId].forEach((service) => {
```

frontend/reports/client-zipcode/client-zipcode-results.component.tsx
```diff
@@ -97,4 +99,31 @@ export default function ClientZipcodeResults(props) {
       />
     </>
   );
+
+  function getCsvOptions(): Promise<CsvOptions> {
+    const countyZips = [];
+    Object.keys(data.zipsByCounty).map((county) => {
```
joeldenning, line 105:
```suggestion
    Object.keys(data.zipsByCounty).forEach((county) => {
```

joeldenning (approved):
Looks good.

PR #809, opened 2021-05-04
interactions by service & revenue by service done

zaratejavier:
image
image

### DIFF
--- a/frontend/reports/interactions-by-service/interactions-by-service-results.component.tsx
+++ b/frontend/reports/interactions-by-service/interactions-by-service-results.component.tsx
@@ -171,7 +171,7 @@ export default function InteractionsByService(props) {
       Service: "---",
       "Client Count": data.grandTotal.numClients.toLocaleString(),
       "Interaction Count": data.grandTotal.numInteractions,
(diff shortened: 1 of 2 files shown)


frontend/reports/interactions-by-service/interactions-by-service-results.component.tsx
```diff
@@ -162,4 +164,72 @@ export default function InteractionsByService(props) {
...
     </>
   );
+
+  function getCsvOptions(data): Promise<CsvOptions> {
+    const allPrograms = {
+      Program: "All programs",
+      Service: "---",
+      "Client Count": data.grandTotal.numClients.toLocaleString(),
+      "Interaction Count": data.grandTotal.numInteractions,
+      "Interaction Hours": `"${formatDuration(data.grandTotal.totalDuration)}"`,
```
joeldenning, line 174: It might be best to not format the duration, since Excel is capable of treating `00:30` as a duration that it can do math equations on, but can't do the same for `0 hours, 30 minutes`

Similar change should be made to every other place in this file that calls formatDuration

```suggestion
      "Interaction Hours": `"${data.grandTotal.totalDuration}"`,
```

PR #851, opened 2021-11-13
Issue 746

nick-nack-attack:
- Modify events report
- Add the gender of attendees
- Add leads that attended along with their current status (if converted to a client)
- Created a new endpoint to handle getting leads from a certain date range
- small refactoring of a ternary operator


joeldenning:
I'd rather not work with you, as discussed in Slack, so I'm closing this.

PR #895, opened 2025-02-03
Mysql2 fixes

joeldenning:


backend/apis/clients/client-logs/get-activity-logs.api.js
```diff
@@ -49,7 +49,7 @@ app.get("/api/clients/:clientId/logs", (req, res, next) => {
     }
 
     logs.forEach((log) => {
-      log.tags = JSON.parse(log.tags)
+      log.tags = log.tags
```
joeldenning, line 52: In the mysql2 package, `json_arrayagg(json_object())` comes back as a js object rather than a string, like it did in the mysql package

backend/apis/invoices/get-invoice-pdf.api.js
```diff
@@ -237,7 +237,7 @@ app.get("/api/invoices/:invoiceId/pdfs", (req, res) => {
           );
 
           const amount = `$${
-            invoice.totalCharged ? invoice.totalCharged.toFixed(2) : "0.00"
+            invoice.totalCharged ? invoice.totalCharged : "0.00"
```
joeldenning, line 240: In `mysql2` package, a column with `decimal(15, 2)` comes back as a string rather than a number

